PR #1, opened 2021-06-29
BXC-3170 - Project initialization and fields commands

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3170

* Adds command for initializing new migration projects
    * Creates project directory if needed, with structure and config
    * Retrieves and stores CDM field info for the collection. Fails if cannot find CDM collection.
* Adds command for validating cdm_fields.json file
* Adds classes to support creation/management of MigrationProjects and cdm field info


src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmFieldService.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * Get the URL for retrieving field info for the given collection
+     * @param cdmBaseUri
+     * @param collectionId
+     * @return
+     */
+    public static String getFieldInfoUrl(String collectionId) {
+        return "dmwebservices/index.php?q=dmGetCollectionFieldInfo/" + collectionId + "/json";
```
lfarrell, line 81: Don't know if things ever improved on the CDM side, but getting JSON sometimes returned a less complete response than getting XML.

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         repository: UNC-Libraries/Carolina-Digital-Repository.git
-        ref: 359e459ad6c3a02b7aa68d7bfab5a90ad8cfb38f
+        ref: f580221852b673e1af442859d229393e80518b13
```
lfarrell, line 21: Will this need to be updated every time a boxc pr is merged?

src/test/java/edu/unc/lib/boxc/migration/cdm/AbstractCommandIT.java
```diff
@@ -0,0 +1,89 @@
...
+            System.setOut(originalOut);
+            log.error(output);
+            fail("Expected command to result in success: " + String.join(" ", args));
+        }
+    }
+
+    protected void executeExpectFailure(String[] args) {
+        int result = migrationCommand.execute(args);
+        output = out.toString();
+        if (result != 1) {
```
lfarrell, line 83: Should this be `result == 0`, in case a different code is returned?

src/test/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactoryTest.java
```diff
@@ -0,0 +1,235 @@
...
+                projectCreated.getProjectProperties().getCreatedDate(),
+                projectLoaded.getProjectProperties().getCreatedDate());
+    }
+
+    private void assertPropertiesFilePopulated(MigrationProject project, String expName, String expCollId)
+            throws IOException {
+        Path propertiesPath = project.getProjectPropertiesPath();
+        assertTrue("Properties object must exist", Files.exists(propertiesPath));
+        MigrationProjectProperties properties = ProjectPropertiesSerialization.read(propertiesPath);
+        assertPropertiesSet(properties, expName, expCollId);
```
lfarrell, line 222: I suppose it doesn't matter too much, but should this be `assertReturnedPropertiesPopulate`?

lfarrell (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmFieldService.java
bbpennel: The responses seem to be complete for getting collection field info, and it is the same approach being used in the NCDHC migration scripts, so I'm guessing it will be consistent. But i did see the same issue you're talking about when it comes to getting item records as json, it would only give me up to 5 fields.

.github/workflows/build.yml
bbpennel: It could be set to `main`, but if it needs to reference something that's not merged yet we'd need to specify a commit. It would just need to be updated when it needs a more recent version of boxc, which would unfortunately be for most PRs. I'm not sure how else to handle it.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactoryTest.java
bbpennel: No, `assertReturnedPropertiesPopulated` checks the properties directly held by a MigrationProject instance, while `assertPropertiesFilePopulated` checks the properties held in the project.json file persisted on disk. `assertPropertiesSet` checks the values of properties regardless of the source, which is why its used by both of the other methods.

src/test/java/edu/unc/lib/boxc/migration/cdm/AbstractCommandIT.java
bbpennel: Updated

.github/workflows/build.yml
lfarrell: Do you think this would work, https://github.com/actions/checkout#checkout-head?

bbpennel:
Hmm, wouldn't HEAD be roughly the same as doing main? It'd require things
to be merged in main, and tracks either the most recent or semi-recent
merges into main. In this case I could just change it to main, and switch
off main only when necessary.

PR #2, opened 2021-07-07
BXC-3171 - Unpaged record exports

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3171
* Adds command and service for exporting CDM records as XML based on field config for small collections that don't requiring exporting in pages

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmExportCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmExportCommand.java
@@ -1,0 +1,119 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/CdmExportCommand.java
```diff
@@ -0,0 +1,119 @@
...
+            defaultValue = "${env:CDM_BASE_URL:-http://localhost:82/}")
+    private String cdmBaseUri;
+    @Option(names = { "-u", "--cdm-user"},
+            description = {"User name for CDM requests.",
+                    "Defaults to current user: ${DEFAULT-VALUE}"},
+            defaultValue = "${sys:user.name}")
+    private String cdmUsername;
+    @Option(names = {"-p", "--cdm-password"},
+            description = "Password for CDM requests. Required.",
+            arity = "0..1",
```
lfarrell, line 63: So, this can take 0 or 1 argument?

lfarrell (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/CdmExportCommand.java
bbpennel: Sort of, this lets you either provide your password directly as the value of the option, or if you don't include a value it will prompt you to type in your password.

PR #3, opened 2021-07-09
BXC-3173 - Index of flat collections

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3173

* Adds a command and services to produce a sqlite database containing exported records from CDM, using the retrieved field information to produce the schema
* Adds properties to keep track of export and indexing state.

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/model/CdmFieldInfo.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/model/CdmFieldInfo.java
@@ -17,7 +17,9 @@ package edu.unc.lib.boxc.migration.cdm.model;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -30,6 +32,8 @@ import org.apache.jena.ext.com.google.common.collect.Streams;
  */
 public class CdmFieldInfo {
     public static final String CDM_ID = "cdmid";
+    public static final Set<String> IGNORE_FIELDS = new HashSet<>(Arrays.asList(
+            "dmoclcno", "dmcreated", "dmmodified", "dmrecord"));
     public static final List<String> RESERVED_FIELDS = Arrays.asList(
             CDM_ID, "cdmcreated", "cdmmodified", "cdmfile", "cdmpath");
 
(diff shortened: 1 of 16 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/model/CdmFieldInfo.java
```diff
@@ -16,14 +16,27 @@
...
+
+import org.apache.jena.ext.com.google.common.collect.Streams;
 
 /**
  * Information describing the fields present in a CDM collection
  *
  * @author bbpennel
  */
 public class CdmFieldInfo {
+    public static final String CDM_ID = "cdmid";
```
lfarrell, line 34: Should this be its own constant if no other field is?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    private String makeInsertTemplate(List<String> exportFields) {
+        return "insert into " + TB_NAME + " values ("
+                + exportFields.stream().map(f -> "?").collect(Collectors.joining(","))
+                + ")";
+    }
+
+    private void indexDocument(Document doc, Connection conn, String insertTemplate, List<String> exportFields)
```
lfarrell, line 110: Just a general question. Is there only one table?

lfarrell (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
bbpennel: yes, currently its a single table

src/main/java/edu/unc/lib/boxc/migration/cdm/model/CdmFieldInfo.java
bbpennel: I made a this one into a constant because I needed to reference the field in another location down here https://github.com/UNC-Libraries/cdm-to-boxc-migration-util/pull/3/files/c22414f4f71187771a869e4762ccfece64ac086b
I didn't need to refer to any of the other reserved fields by name anywhere else so far, but i would probably make constants for them if i end up needing to do so

lfarrell (approved):

PR #4, opened 2021-07-15
BXC-3208 - Basic destination mapping

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3208

* Adds command for generating basic one destination mapping files, and some infrastructure for later more complicated cases

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DestinationsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DestinationsService.java
@@ -1,0 +1,160 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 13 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/DestinationsService.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private void assertProjectStateValid() {
+        if (project.getProjectProperties().getIndexedDate() == null) {
+            throw new InvalidProjectStateException("Project must be indexed prior to generating destinations");
+        }
+    }
+
+    private void assertValidDestination(String destination) {
+        if (destination == null) {
```
lfarrell, line 96: Should this case throw an exception too?

lfarrell (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/DestinationsService.java
bbpennel: Updated

PR #11, opened 2021-10-01
BXC-3203 - Update Access Surrogate datastream name

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3203

Due to renaming of constants from boxc


bbpennel:
Tests will fail until https://github.com/UNC-Libraries/box-c/pull/1298 is merged

PR #14, opened 2021-11-03
Catch output to prevent intellij test failures

bbpennel:
Add helper for consuming output from tests, use it for some of the dry run tests since otherwise intellij can't run them


sharonluong (approved):
Looks good. I tested running the GroupMappingServiceTest and the SourceFileServiceTest and they both passed fully.

PR #17, opened 2021-11-15
BXC-3228 - Compound object support

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3228

* Adds support for migration of compound objects, which are predefined multi-file works in CDM
    * Supports migration of compound object children
    * Compound children don't have the reserved CDM fields that their parent objects do (like cdmid, last modified, etc), so some refactoring of how lists of fields are calculated to account for this. Made some related variables into instance variables instead of method params to cut down on redundancy
* Exclude grouped works from description generation command, since they cannot have descriptions associated with them and cause errors when generating SIPs
* Test helper updates to support testing with other example collections


sharonluong (changes requested):
Just two requests for clarifying comments/javadocs

PR #21, opened 2021-12-20
Bxc 3371 redirect mapping indexing service

sharonluong:
This PR satisfies [BXC-3371](https://jira.lib.unc.edu/browse/BXC-3371) which creates an indexing service for the redirect mapping information. The service reads the redirect-mapping CSV in the project directory and inserts values into the redirect_mapping table in the chomping_block database.

In this particular iteration, the service is inserting into a sqlite DB that the RedirectMappingIndexServiceTest accesses to check assertions. Hooking up into the actual DBs on the servers will happen in  [BXC-3372](https://jira.lib.unc.edu/browse/BXC-3372)


src/main/java/edu/unc/lib/boxc/migration/cdm/model/MigrationProject.java
```diff
@@ -36,6 +36,7 @@
     public static final String GROUP_MAPPING_FILENAME = "group_mappings.csv";
     public static final String SIPS_DIRNAME = "sips";
     public static final String REDIRECT_MAPPING_FILENAME = "redirect_mappings.csv";
+    public static final String REDIRECT_MAPPING_INDEX_FILENAME = "redirect_mapping_index.db";
```
bbpennel, line 39: Since this filepath is only going to be used for tests, I would prefer this constant and the path construction be moved into a test class. You could move them into SipServiceHelper.

If you do this, then you will need to add a way to pass the database connection string into RedirectMappingIndexService, which you could do by adding a setter (`setConnectionString`) to the service and then having the test call the setter and pass in value`"jdbc:sqlite:" + testHelper.getRedirectMappingIndexPath()`

You'll need to do something like this eventually either way since there will be two database types, so if you'd like we could defer this to the next ticket.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -0,0 +1,160 @@
...
+        try {
+            indexService.indexMapping();
+        } catch (InvalidProjectStateException e) {
+            assertTrue("Unexpected message: " + e.getMessage(),
+                    e.getMessage().contains("Must submit the collection prior to indexing"));
+        }
+    }
+
+    @Test
+    public void redirectMappingTableIsPopulatedWithRightNumberOfRows() throws Exception {
```
bbpennel, line 100: You could merge this test with the next one, they are both verifying that the same set of redirects mapping made it into the database, using two different metrics to verify it. I generally try to base tests around the state/condition that is being tested rather than the way the outcome is validated. If the test is getting super long then it can make sense to split it up or use private helper methods to improve readability, but i think in this case it'd be okay to merge.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -0,0 +1,160 @@
...
+            rs.next();
+            assertEquals("Incorrect number of rows in database", 3, rs.getInt(1));
+        } finally {
+            CdmIndexService.closeDbConnection(conn);
+        }
+    }
+
+    @Test
+    public void  redirectMappingIndexPopulatesColumnsCorrectly() throws Exception {
+        ArrayList<String> row1 = new ArrayList<>();
```
bbpennel, line 117: Just a java thing, but you typically want to declare variables by their interface rather than their implementation (`ArrayList` is an implementation of the `List` interface).
```suggestion
        List<String> row1 = new ArrayList<>();
```
In java 11 there are other ways to handle this, but we can look at that when we are able to make the jump.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -0,0 +1,160 @@
...
+            rs.next();
+            assertEquals("cdm_collection_id value isn't accurate", row1.get(0), rs.getString("cdm_collection_id"));
+            assertEquals("cdm_object_id value isn't accurate", row1.get(1), rs.getString("cdm_object_id"));
+            assertEquals("boxc_work_id value isn't accurate", row1.get(2), rs.getString("boxc_work_id"));
+            assertEquals("boxc_file_id value isn't accurate", row1.get(3), rs.getString("boxc_file_id"));
+        } finally {
+            CdmIndexService.closeDbConnection(conn);
+        }
+    }
+
```
bbpennel, line 150: I would suggest having one more test to verify that compound objects work (cases where we have an entry without a File id) and that nulls are getting set, since if not that will mess up the queries on the redirect end. This can be setup with the following:
```
        testHelper.indexExportData(Paths.get("src/test/resources/keepsakes_fields.csv"), "export_compounds.xml");
        testHelper.getDescriptionsService().generateDocuments(true);
        testHelper.getDescriptionsService().expandDescriptions();
        testHelper.populateSourceFiles("nccg_ck_09.tif", "nccg_ck_1042-22_v1.tif",
                "nccg_ck_1042-22_v2.tif", "nccg_ck_549-4_v1.tif", "nccg_ck_549-4_v2.tif");
```
This is stolen from `StatusCommandIT`

src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
```diff
@@ -0,0 +1,104 @@
...
+ * @author snluong
+ */
+public class RedirectMappingIndexService {
+    private final MigrationProject project;
+
+    public RedirectMappingIndexService(MigrationProject project) {
+        this.project = project;
+    }
+
+    public void indexMapping() {
```
bbpennel, line 46: I know it's not going to say a lot, but please add javadoc to this method to briefly describe what the method does. Typically we want to have javadoc for all public methods that aren't just boilerplate like getters and setters.

bbpennel (changes requested):
Looks good! I added suggestions for one more test and some cleanup

src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
```diff
@@ -0,0 +1,104 @@
...
+                    .withHeader(RedirectMappingService.CSV_HEADERS)
+                    .withTrim());
+
+            for (CSVRecord originalRecord : originalParser) {
+                String cdm_collection_id = originalRecord.get(0);
+                String cdm_object_id = originalRecord.get(1);
+                String boxc_work_id = originalRecord.get(2);
+                String boxc_file_id = originalRecord.get(3);
+
+                String query = " insert into redirect_mappings " +
```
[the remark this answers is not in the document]
sharonluong: sounds good

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
sharonluong: Can you explain this a bit more? I don't mind doing it but I don't understand the difference? You declare it as a List with the variable name but then the instantiation is an ArrayList?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
sharonluong: ok!

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
sharonluong: ok, got it

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
bbpennel: The intent is to decouple code from details of the implementation, so that you could swap in other implementations of lists later if need be. This stackoverflow thread talks about it:
https://stackoverflow.com/questions/9852831/polymorphism-why-use-list-list-new-arraylist-instead-of-arraylist-list-n

Generally you only declare a variable as the implementation type if you need functionality specific to that implementation. It seems a bit abstract, but it matters more in some places than others, like if you have a data model that you need to implement for two different back ends (maybe one is database driven and the other is from local files), and have code that uses the data model but doesn't care which implementation is used.

It doesn't matter that much here, since ArrayList only has 2-3 methods not available to the List interface, so mainly its a best practice that you'll see all over the place in java code.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
sharonluong: Okay, I've added a compound objects test that makes sure that the boxc_work_id is always populated. Let me know if I could have implemented that a better way!

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -0,0 +1,195 @@
...
+                "created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, " +
+                "UNIQUE (cdm_collection_id, cdm_object_id)" +
+                ")");
+        CdmIndexService.closeDbConnection(conn);
+    }
+
+    @Test
+    public void indexingDoesNotHappenIfCollectionIsNotSubmitted() {
+        try {
+            indexService.indexMapping();
```
bbpennel, line 87: after this line, add `fail();` (its another junit assertion method), otherwise the test will pass even if no exception is thrown

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -0,0 +1,195 @@
...
+        addSipsSubmitted();
+        indexService.indexMapping();
+
+        try {
+            Statement stmt = conn.createStatement();
+            ResultSet count = stmt.executeQuery("select count(*) from redirect_mappings");
+            count.next();
+            assertEquals("Incorrect number of rows in database", 7, count.getInt(1));
+
+            ResultSet rs = stmt.executeQuery("select boxc_work_id from redirect_mappings where boxc_work_id is null");
```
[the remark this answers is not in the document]
sharonluong: yeahhhh I was wondering if that was the better test but I went with your suggestion in slack:
`it is important to know that the index and the csv end up with the same work id at least once, but one approach for the current test could be to simply require that the work id not be null or empty`

i'll try checking for the null file IDs instead?

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -140,8 +142,11 @@ public void redirectMappingIndexPopulatesTableCorrectly() throws Exception {
     }
 
     @Test
-    public void workIdsArePopulatedInRedirectMappingIndexForCompoundObjects() throws Exception {
-        List<String> nulls = new ArrayList<>();
+    public void tableIsPopulatedCorrectlyInRedirectMappingIndexForCompoundObjects() throws Exception {
+        List<String> cdm_object_ids = new ArrayList<>();
+        List<String> expected_ids = new ArrayList<>();
+        expected_ids.add("604");
+        expected_ids.add("607");
```
bbpennel, line 149: you don't have to change this, but just so you know if you want to initialize a list with a few elements and don't need to change the list later, you can do this
```suggestion
        List<String> expected_ids = Arrays.asList("604", "607");
```

bbpennel (approved):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
sharonluong: Ugh i was looking for how to do that but didn't see anything. thanks for the tip!

PR #24, opened 2022-01-12
Bxc 3372 connect redirect service to dbs

sharonluong:
https://jira.lib.unc.edu/browse/BXC-3372

This PR does a few things:
- builds the command class for indexing the redirect mapping csv to the DB
- code that takes the file path of the properties file which has the DB credentials and uses that to generate a connection string to the DB
- differentiates between using MySQL on the actual servers and sqlite for the tests
- adds IT for command class and one more RedirectMappingService test that builds the connection string

There is a second part TBD which will be on the boxc-ansible repo that takes care of the properties files on each server.

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
@@ -49,16 +49,24 @@ public class IndexRedirectCommand implements Callable<Integer> {
     @Override
     public Integer call() throws Exception {
         try {
+            validate();
             Path currentPath = parentCommand.getWorkingDirectory();
             project = MigrationProjectFactory.loadMigrationProject(currentPath);
             indexService = new RedirectMappingIndexService(project);
             indexService.setRedirectDbConnectionPath(redirectDbConnectionPath);
+            indexService.init();
 
             indexService.indexMapping();
             return 0;
         } catch (MigrationException e) {
             outputLogger.info("Failed to index redirect mapping: {}", e.getMessage());
             return 1;
+        }
+    }
+
+    private void validate() {
+        if (redirectDbConnectionPath == null) {
+            throw new MigrationException("The DB connection path must be included");
         }
     }
 }
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
@@ -57,6 +57,7 @@ public class IndexRedirectCommand implements Callable<Integer> {
             indexService.init();
 
             indexService.indexMapping();
+            outputLogger.info("Redirect mapping indexing completed. Yay!");
             return 0;
         } catch (MigrationException e) {
             outputLogger.info("Failed to index redirect mapping: {}", e.getMessage());
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
@@ -57,7 +57,7 @@ public class IndexRedirectCommand implements Callable<Integer> {
             indexService.indexMapping();
             return 0;
         } catch (MigrationException e) {
-            outputLogger.info("Failed to export project: {}", e.getMessage());
+            outputLogger.info("Failed to index redirect mapping: {}", e.getMessage());
             return 1;
         }
     }
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/IndexRedirectCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public Integer call() throws Exception {
+        try {
+            Path currentPath = parentCommand.getWorkingDirectory();
+            project = MigrationProjectFactory.loadMigrationProject(currentPath);
+            indexService = new RedirectMappingIndexService(project);
+            indexService.setRedirectDbConnectionPath(redirectDbConnectionPath);
+
+            indexService.indexMapping();
```
bbpennel, line 57: I'd suggest adding an `outputLogger.info` statement after the indexing completes to say that it successfully completed, so users will get a sense that something happened. You could also then test for that in your integration test.

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
```diff
@@ -0,0 +1,91 @@
...
+        return project.getProjectPath().resolve(REDIRECT_MAPPING_INDEX_FILENAME);
+    }
+
+    public SipGenerationOptions makeOptions() {
+        SipGenerationOptions options = new SipGenerationOptions();
+        options.setUsername(USERNAME);
+        return options;
+    }
+
+    public Path createRedirectDbConnectionPropertiesFile(TemporaryFolder tempFolder, String dbType) throws IOException {
```
bbpennel, line 81: Feel free to ignore, but given that this method is in the `RedirectMappingHelper` I would say you don't really need to include `Redirect` in the method name

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
```diff
@@ -0,0 +1,91 @@
...
+public class RedirectMappingHelper {
+    private static final String USERNAME = "migr_user";
+    public static final String REDIRECT_MAPPING_INDEX_FILENAME = "redirect_mapping_index.db";
+    private final MigrationProject project;
+
+    public RedirectMappingHelper(MigrationProject project) {
+        this.project = project;
+    }
+
+    public void createRedirectMappingsTableInDb(Path redirectMappingIndexPath) {
```
bbpennel, line 46: Could drop the `redirectMappingIndexPath` parameter since it is always being passed the result from `getRedirectMappingIndexPath`, and call that method directly.

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public SipGenerationOptions makeOptions() {
+        SipGenerationOptions options = new SipGenerationOptions();
+        options.setUsername(USERNAME);
+        return options;
+    }
+
+    public Path createRedirectDbConnectionPropertiesFile(TemporaryFolder tempFolder, String dbType) throws IOException {
+        String mysqlString = "db_type=mysql\ndb_host=localhost\ndb_user=root\ndb_password=password";
```
bbpennel, line 82: Up to you which you prefer, but wanted to mention that rather than using string concatenation for building the properties file, it looks like you could create and serialize a properties object like in section 1 of this document https://mkyong.com/java/java-properties-file-examples/

src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
```diff
@@ -97,15 +100,50 @@ private void assertCollectionSubmitted() {
...
+    }
+
+    public Properties getConnectionProperties() throws IOException {
+        InputStream inputStream = Files.newInputStream(redirectDbConnectionPath);
+        Properties props = new Properties();
+        props.load(inputStream);
+        return props;
+    }
+
+    public String generateConnectionString(Properties props) {
```
bbpennel, line 135: Since this is a public method which does a fair amount of work, could you add javadoc to this method?

bbpennel (changes requested):
Pending testing on the server, this looks good to me. Most of my comments are optional suggestions

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
sharonluong: thanks! decided to keep it as is because the current implementation is a bit simpler and less lines.

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final String USERNAME = "migr_user";
+    public static final String REDIRECT_MAPPING_INDEX_FILENAME = "redirect_mapping_index.db";
+    private final MigrationProject project;
+
+    public RedirectMappingHelper(MigrationProject project) {
+        this.project = project;
+    }
+
+    public void createRedirectMappingsTableInDb() {
+        String connectionString = "jdbc:sqlite:" + getRedirectMappingIndexPath();
```
bbpennel, line 50: for what its worth, it seems like you could use `RedirectMappingIndexService.openDbConnection` for initializing the database connection, which would make sure they work the same way, although it would only save you a few lines here

bbpennel (approved):

src/test/java/edu/unc/lib/boxc/migration/cdm/test/RedirectMappingHelper.java
sharonluong: yeah that's true. i'll change it after my meeting at 2

sharonluong: upon further thought, i'll just keep it simple as is =)

PR #29, opened 2022-02-03
Bxc 3379 set properties

krwong:
[https://jira.lib.unc.edu/browse/BXC-3379](https://jira.lib.unc.edu/browse/BXC-3379)
- `MigrationProjectProperties`: add hookId and collectionNumber to project properties
- `ProjectPropertiesService`: service to list configurable properties and set/unset those properties
- `ProjectPropertiesCommand`: add list, set, and unset commands
- Remaining changes included service and command tests.
- Other notes: Currently, only hookId and collectionNumber can be manually set/unset. You can only set one property at a time. You can unset multiple properties at the same time.


src/main/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesService.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private MigrationProject project;
+
+    /**
+     * List project properties that can be manually set/unset
+     * @throws IOException
+     */
+    public void getProjectProperties() throws IOException {
+        Map<String, String> map = new HashMap<String, String>();
+        map.put("hookId", project.getProjectProperties().getHookId());
```
bbpennel, line 43: Please make constants for the hookId and collectionNumber field names, then use those constants in all the methods in this class.

Also, I wanted to mention that it is possible to construct the names of methods using strings and invoke them programmatically, versus adding cases for each field. This can be done using reflection:
https://www.baeldung.com/java-method-reflection

```
Method getterMethod = MigrationProjectProperties.class.getMethod("getHookId");
```
It would probably be fine to do this in another ticket focused on making more fields listable though since it'd be a pretty significant change.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesService.java
```diff
@@ -0,0 +1,92 @@
...
+public class ProjectPropertiesService {
+
+    private MigrationProject project;
+
+    /**
+     * List project properties that can be manually set/unset
+     * @throws IOException
+     */
+    public void getProjectProperties() throws IOException {
+        Map<String, String> map = new HashMap<String, String>();
```
bbpennel, line 42: If intellij has been set to use java 11 as the language version, you could switch this to `var map = new HashMap<String, String>();`, or could use the shorthand `Map<String, String> map = new HashMap<>();`

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesService.java
```diff
@@ -0,0 +1,92 @@
...
+        map.put("hookId", project.getProjectProperties().getHookId());
+        map.put("collectionNumber", project.getProjectProperties().getCollectionNumber());
+        Set<Map.Entry<String, String>> set = map.entrySet();
+        List<Map.Entry<String, String>> configuredProperties = new ArrayList<>(set);
```
bbpennel, line 46: There are a few methods that would allow us to iterate directly against the map or entry set, which could save some steps:
https://stackoverflow.com/questions/4234985/how-to-for-each-the-hashmap
Although based on my other comment we might want to do that in the command class rather than here.

src/test/java/edu/unc/lib/boxc/migration/cdm/ProjectPropertiesCommandIT.java
```diff
@@ -0,0 +1,115 @@
...
+    @Rule
+    public final TemporaryFolder tmpFolder = new TemporaryFolder();
+
+    private MigrationProject project;
+    private ProjectPropertiesService propertiesService;
+
+    @Before
+    public void setup() throws Exception {
+        project = MigrationProjectFactory.createMigrationProject(
+                baseDir, PROJECT_NAME, null,"user");
```
bbpennel, line 42:
```suggestion
                baseDir, PROJECT_NAME, null, "user");
```

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesServiceTest.java
```diff
@@ -0,0 +1,119 @@
...
+            fail();
+        } catch (MigrationException e) {
+            assertTrue(e.getMessage().contains("Invalid field/value input"));
+        }
+    }
+
+    @Test
+    public void setProjectPropertiesNullValueTest() throws Exception {
+        try {
+            service.setProperties("hookId",null);
```
bbpennel, line 88:
```suggestion
            service.setProperties("hookId", null);
```

bbpennel (changes requested):
This looks good over all, I've left some suggestions

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesService.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String HOOK_ID = "hookId";
+    private static final String COLLECTION_NUMBER = "collectionNumber";
+
+    /**
+     * List project properties that can be manually set/unset
+     * @throws IOException
+     */
+    public Map<String, String> getProjectProperties() throws IOException {
+        Map<String, String> projectProperties = new HashMap<>();
+        projectProperties.put(HOOK_ID, project.getProjectProperties().getCollectionNumber());
```
bbpennel, line 43: should be getHookId

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ProjectPropertiesService.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Map;
+
+/**
+ * Service for manually updating the project properties
+ * @author krwong
+ */
+public class ProjectPropertiesService {
+    private MigrationProject project;
+
+    private static final String HOOK_ID = "hookId";
```
bbpennel, line 34: If you make these constants public then you could also use them in tests or other places where we need to reference these fields by their identifier.

bbpennel (changes requested):

bbpennel (approved):

PR #30, opened 2022-02-04
WIP: BXC-3436 first pass at logic for adding collection-specific row to re…

sharonluong:
…direct mapping csv

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingService.java
@@ -15,14 +15,22 @@
  */
 package edu.unc.lib.boxc.migration.cdm.services;
 
+import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
```diff
@@ -83,7 +83,11 @@ public void indexMapping() {
 
                 PreparedStatement preparedStatement = conn.prepareStatement(INSERT_STATEMENT);
                 preparedStatement.setString(1, cdm_collection_id);
-                preparedStatement.setString(2, cdm_object_id);
+                if (cdm_object_id.isEmpty()) {
```
bbpennel, line 86: Its sort of not related to this PR, but I'm realizing this class has a bunch of variables using snake case instead of camel case. We could address that now, or do so in a follow up PR.

bbpennel (approved):
I left one comment, but I'm okay with merging now and addressing it separately

src/main/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexService.java
sharonluong: ah i always get confused on which is proper format for which language. so variable names are camelCase?

bbpennel (approved):

PR #32, opened 2022-02-22
Bxc 3337 record finding aid

krwong:
[https://jira.lib.unc.edu/browse/BXC-3337](https://jira.lib.unc.edu/browse/BXC-3337)

- `FindingAidService`: service to check for and record hookId and collectionNumber project properties
- `InitializeProjectCommand`: add FindingAidService to command
- Remaining changes include service tests, command tests, and a test resource.

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright 2008 The University of North Carolina at Chapel Hill
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.unc.lib.boxc.migration.cdm.services;
+
+import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
+
+import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
+import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
+import edu.unc.lib.boxc.migration.cdm.util.ProjectPropertiesSerialization;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Service for recording a collection's finding aid
+ * @author krwong
+ */
+public class FindingAidService {
+    public static final String CONTRI_FIELD = "contri";
+    public static final String DESCRI_FIELD = "descri";
+    public static final String HOOK_ID = "hook id";
+    public static final String COLLECTION_NUMBER = "collection number";
+
+    private MigrationProject project;
+    private CdmFieldService fieldService;
+
+    public void recordFindingAid() throws Exception {
+        fieldService.validateFieldsFile(project);
+        CdmFieldInfo fieldInfo = fieldService.loadFieldsFromProject(project);
+        Map<String, String> fields = fieldInfo.getFields().stream()
+                .filter(f -> !f.getSkipExport())
+                .collect(Collectors.toMap(CdmFieldInfo.CdmFieldEntry::getExportAs,
+                        CdmFieldInfo.CdmFieldEntry::getDescription));
+
+        for (Map.Entry<String, String> entry : fields.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
```diff
@@ -0,0 +1,70 @@
...
+public class FindingAidService {
+    public static final String CONTRI_FIELD = "contri";
+    public static final String DESCRI_FIELD = "descri";
+    public static final String HOOK_ID = "hookid";
+    public static final String COLLECTION_NUMBER = "collection number";
+
+    private MigrationProject project;
+    private CdmFieldService fieldService;
+
+    public void recordFindingAid() throws Exception {
```
bbpennel, line 40: Please add some javadoc for this method

src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
```diff
@@ -0,0 +1,70 @@
...
+                .collect(Collectors.toMap(CdmFieldInfo.CdmFieldEntry::getNickName,
+                        CdmFieldInfo.CdmFieldEntry::getDescription));
+
+        for (Map.Entry<String, String> entry : fields.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            if (key.equalsIgnoreCase(CONTRI_FIELD) && value.equalsIgnoreCase(HOOK_ID)) {
+                project.getProjectProperties().setHookId(key);
+                ProjectPropertiesSerialization.write(project);
+                outputLogger.info("HookId was set. Use 'config list' to view the project property.");
```
bbpennel, line 54: For this message, it would be good to use the  constant ProjectPropertiesService.HOOK_ID instead of directly putting "HookId" in the message, since then the user will know the name of the exact property that was set.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
```diff
@@ -0,0 +1,70 @@
...
+            } else if (key.equalsIgnoreCase(DESCRI_FIELD) && value.equalsIgnoreCase(COLLECTION_NUMBER)) {
+                project.getProjectProperties().setCollectionNumber(key);
+                ProjectPropertiesSerialization.write(project);
+                outputLogger.info("CollectionNumber was set. Use 'config list' to view the project property.");
```
bbpennel, line 58: Same suggestion as with hookid

src/test/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+            service.recordFindingAid();
+            fail();
+        } catch (MigrationException e) {
+            //Expected
+        }
+    }
+
+    @Test
+    public void noFindingAidToRecordTest() throws Exception {
+        populateFieldInfo();
```
bbpennel, line 78: I'd suggest either removing the private method and putting its body directly here (since this is the only usage), or changing the private method to accept the path you want to copy over as a parameter and using it for `recordFindingAidTest` as well.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void recordFindingAidTest() throws Exception {
+        Files.copy(Paths.get("src/test/resources/findingaid_fields.csv"), project.getFieldsPath());
+        service.recordFindingAid();
+
+        assertEquals(CONTRI_FIELD, project.getProjectProperties().getHookId());
+        assertEquals(DESCRI_FIELD, project.getProjectProperties().getCollectionNumber());
+    }
+
+    @Test
+    public void recordFindingAidFailTest() throws Exception {
```
bbpennel, line 67: Is this method testing cases where the service is executed without getting a field mapping? If so, might be clearer to name it `recordFindingAidNoFieldCsv` to describe the test case.

I often just list what code is being tested and what the conditions are, although it can also be good to list what the expected outcome is (to include all three I'd probably name this `recordFindingAidNoFieldCsvFail`). Its mostly up to you what seems clearest, just suggesting some options.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidService.java
```diff
@@ -51,11 +56,11 @@ public void recordFindingAid() throws Exception {
             if (key.equalsIgnoreCase(CONTRI_FIELD) && value.equalsIgnoreCase(HOOK_ID)) {
                 project.getProjectProperties().setHookId(key);
                 ProjectPropertiesSerialization.write(project);
-                outputLogger.info("HookId was set. Use 'config list' to view the project property.");
+                outputLogger.info(HOOK_ID + " was set. Use 'config list' to view the project property.");
```
bbpennel, line 59: Sorry, I meant use `ProjectPropertiesService.HOOK_ID`, since that is the name of the property the user would have to change using the config command if they disagreed with the automatic assignment that took place. Its a little different from the constant from this file (which is the description of the field itself). It might be less confusing if you renamed the constants declared in this class to `HOOK_ID_FIELD_DESC` and `COLLECTION_NUMBER_FIELD_DESC`

PR #33, opened 2022-03-01
Display what the error is if indexing fails, and add some debug

bbpennel:


sharonluong (approved):
thanks for adding logging!

PR #35, opened 2022-03-02
BXC-3490 - Skip missing fields, catch missing records

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3490
* If a configured field is no present for a record or a compound child, then store an empty string for that field. This is to account for some admin fields in cdm not being returned in exports, particularly for compound children
* Change default page size for exports, since CDM is losing ids when exporting 1000 at a time for some collections
* Add verification step to ensure that no ids are missing from an export before declaring export a success.

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportService.java
@@ -178,15 +178,15 @@ public class CdmExportService {
         var exportedIds = recordEls.stream()
                 .map(el -> el.getChildTextTrim(CdmFieldInfo.CDM_ID))
                 .collect(Collectors.toList());
-        var missing = CollectionUtils.removeAll(cdmIds, exportedIds);
-        if (!missing.isEmpty()) {
+        var missingIds = CollectionUtils.removeAll(cdmIds, exportedIds);
+        if (!missingIds.isEmpty()) {
             throw new MigrationException("Export failure, document returned by CDM was missing the following records: "
-                + String.join(", ", missing));
+                + String.join(", ", missingIds));
         }
         if (exportedIds.size() != cdmIds.size()) {
-            var extra = CollectionUtils.removeAll(exportedIds, cdmIds);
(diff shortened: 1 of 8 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportService.java
```diff
@@ -155,6 +165,31 @@ private void downloadExport(CdmExportOptions options, Path exportFilePath) throw
...
+        var exportedIds = recordEls.stream()
+                .map(el -> el.getChildTextTrim(CdmFieldInfo.CDM_ID))
+                .collect(Collectors.toList());
+        var missing = CollectionUtils.removeAll(cdmIds, exportedIds);
+        if (!missing.isEmpty()) {
+            throw new MigrationException("Export failure, document returned by CDM was missing the following records: "
+                + String.join(", ", missing));
+        }
+        if (exportedIds.size() != cdmIds.size()) {
+            var extra = CollectionUtils.removeAll(exportedIds, cdmIds);
```
sharonluong, line 187: can this be changed to `extraIds`?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportService.java
```diff
@@ -155,6 +165,31 @@ private void downloadExport(CdmExportOptions options, Path exportFilePath) throw
...
+            doc = builder.build(exportPath.toFile());
+        } catch (JDOMException | IOException e) {
+            throw new MigrationException("Unable to read export file: " + exportPath, e);
+        }
+        Element root = doc.getRootElement();
+        List<Element> recordEls = root.getChildren("record");
+        var exportedIds = recordEls.stream()
+                .map(el -> el.getChildTextTrim(CdmFieldInfo.CDM_ID))
+                .collect(Collectors.toList());
+        var missing = CollectionUtils.removeAll(cdmIds, exportedIds);
```
sharonluong, line 181: can this be changed to `missingIds` to make it a bit clearer?

sharonluong (changes requested):

sharonluong (approved):
looks good!

PR #37, opened 2022-03-10
BXC-3528 - Quote names of fields

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3528
* Fix bug that causes indexing to fail when CDM fields have reserved words for names by quoting field names


src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
```diff
@@ -91,7 +92,7 @@ public void indexExportOneFileTest() throws Exception {
         Connection conn = service.openDbConnection();
         try {
             Statement stmt = conn.createStatement();
-            ResultSet rs = stmt.executeQuery("select " + Strings.join(exportFields, ",")
+            ResultSet rs = stmt.executeQuery("select " + String.join(",", exportFields)
```
sharonluong, line 95: was Strings a valid object type before? what does this change do?

bbpennel: Previously I had accidentally imported an internal class from a library I'd never heard of (`joptsimple.internal.Strings`) which was a dependency of a dependency instead of the standard java `String.join`. See this deletion https://github.com/UNC-Libraries/cdm-to-boxc-migration-util/pull/37/files#diff-d8fd53b88ff022936511261ca94760e6175c3c95efb1ba511028a3bdc4d87303L49

PR #43, opened 2022-05-26
BXC-3679/3680 - Group mapping ids

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3679
https://jira.lib.unc.edu/browse/BXC-3680

* Single member groups will not be added to the group mapping file (they were already being excluded when syncing mappings back to the database)
* Group key is now the matched value with a prefix, and the matched value field has been removed from the CSV
* Matching between MODS files and group generated works is now based on the group key
* Removed assignment of primary objects from works (for multifile works, such as groups, it was assigning every file as a primary object, and we don't need primary objects in the migration)

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DescriptionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DescriptionsService.java
@@ -239,7 +239,10 @@ public class DescriptionsService {
      * @return The path for the individual MODS file
      */
     public Path getExpandedDescriptionFilePath(String cdmId) {
+        // To avoid having individual directories with too many files, MODS files are written to
+        // EXPANDED_FILES_BUCKETS number of bucket directories based on the hash code of the id provided
         String subdir = Integer.toString(Math.abs(cdmId.hashCode()) % EXPANDED_FILES_BUCKETS);
+        // Replace reserved filepath characters with underscores, primarily for grouped work ids
         String filename = cdmId.replaceAll("[: /\\\\<>|&]", "_");
         return project.getExpandedDescriptionsPath().resolve(subdir).resolve(filename + ".xml");
     }
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DescriptionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/DescriptionsService.java
@@ -68,7 +68,7 @@ import edu.unc.lib.boxc.migration.cdm.util.ProjectPropertiesSerialization;
  */
 public class DescriptionsService {
     private static final Logger log = LoggerFactory.getLogger(DescriptionsService.class);
-    private static final int EXPANDED_FILES_PER_DIR = 1024;
+    private static final int EXPANDED_FILES_BUCKETS = 1024;
 
     private static final QName TYPE_NAME = new QName("type");
     private static final QName DISPLAY_LABEL_NAME = new QName("displayLabel");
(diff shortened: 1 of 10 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingServiceTest.java
```diff
@@ -446,11 +417,19 @@ private void assertParentIdsPresent(Connection conn, String... expectedParents)
         while (rs.next()) {
             parentIds.add(rs.getString(1));
         }
-        List<String> expected = Arrays.asList(expectedParents);
+        var expected = Arrays.stream(expectedParents).map(this::asGroupKey).collect(Collectors.toList());
         assertTrue("Expected parent ids " + expected + " but found " + parentIds, parentIds.containsAll(expected));
         assertEquals("Expected parent ids " + expected + " but found " + parentIds, expected.size(), parentIds.size());
     }
 
+    private String asGroupKey(String matchValue) {
```
sharonluong, line 425: I think I kind of like it better without the else. like just the if block and then the return? also i don't quite get what this method is doing from the name...can you explain?

src/test/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingServiceTest.java
```diff
@@ -105,18 +101,16 @@ public void generateSingleRunTest() throws Exception {
...
 
         GroupMappingInfo info = service.loadMappings();
-        String group1Key = info.getGroupKeyByMatchedValue("groupa:group1");
-        String group2Key = info.getGroupKeyByMatchedValue("groupa:group2");
-        assertMappingPresent(info, "25", "groupa:group1", group1Key);
-        assertMappingPresent(info, "26", "groupa:group1", group1Key);
-        assertMappingPresent(info, "27", "groupa:group2", group2Key);
-        assertMappingPresent(info, "28", null, null);
-        assertMappingPresent(info, "29", null, null);
+        assertMappingPresent(info, "25", "groupa:group1");
```
sharonluong, line 104: both this first set of `assertMappingPresent` assertions match the set of assertions on line 152, 173, and 190. they're supposed to be the same from all the different runs, yes? might be nice to DRY that up?

sharonluong (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/DescriptionsService.java
```diff
@@ -239,9 +239,9 @@ private boolean isCdmIdentifier(StartElement element) {
      * @return The path for the individual MODS file
      */
     public Path getExpandedDescriptionFilePath(String cdmId) {
-        int cdmIdNum = Integer.parseInt(cdmId);
-        String subdir = Integer.toString(cdmIdNum / EXPANDED_FILES_PER_DIR);
-        return project.getExpandedDescriptionsPath().resolve(subdir).resolve(cdmId + ".xml");
+        String subdir = Integer.toString(Math.abs(cdmId.hashCode()) % EXPANDED_FILES_BUCKETS);
```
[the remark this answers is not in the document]
bbpennel: Added a comment, its being used to limit the max number of directories the MODS files get distributed to

src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -99,6 +98,22 @@ public void generateMapping(GroupMappingOptions options) throws IOException {
             conn = indexService.openDbConnection();
             Statement stmt = conn.createStatement();
             stmt.setFetchSize(FETCH_SIZE);
+
+            // Return set of all group keys that have at least 2 records in them
+            var groupSet = new HashSet<String>();
+            ResultSet groupRs = stmt.executeQuery("select " + options.getGroupField()
```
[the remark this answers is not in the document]
bbpennel: its just a database resultset, which i've been using the "Rs" naming scheme for so far, and which is short lived, so I'm inclined to leave it. i did rename the Set that the ResultSet is being used to populate though to try to make the contents/usage clearer.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingServiceTest.java
bbpennel: I've added javadoc to the method. Stylistically, I see the branches in the conditional here as being fairly equal rather than an early return, so I think i prefer keeping the if+else, but don't have strong feelings about it

src/test/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingServiceTest.java
bbpennel: added a common assertion method

sharonluong (approved):

PR #44, opened 2022-06-02
BXC-3665 - Set CDM Env during Init

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3665

* CDM environment is now specified by providing a `-e` option during the `init` command
* Partially migrated application configuration over to a json file rather than passing in via environment variables/options, to accomodate the need for providing info about multiple CDM servers
    * the CLIMain parent command provides access to the application configuration for any commands that need it
    * Removed a number of options that are now covered by the config
    * Moved to a static port for mock web server so that the port could be included in test config files
* Some consolidation of common steps in ITs into parent class, in part ot make it easier to provide new config
* Restart export if trying to resume and still in downloading desc.all phase, rather than blocking user from proceeding.

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
@@ -81,7 +81,7 @@ public class MigrationProjectFactory {
         properties.setCreatedDate(Instant.now());
         properties.setName(projectName);
         properties.setCdmCollectionId(collectionId == null ? projectName : collectionId);
-        properties.setCdmEnvironment(cdmEnv);
+        properties.setCdmEnvironmentId(cdmEnv);
         project.setProjectProperties(properties);
         ProjectPropertiesSerialization.write(propertiesPath, properties);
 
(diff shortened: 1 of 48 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/test/CdmEnvironmentHelper.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static Map<String, CdmEnvironment> getTestMapping() {
+        return Map.of(DEFAULT_ENV, getTestEnv());
+    }
+
+    /**
+     * @return CdmEnvironment for a test environment
+     */
+    public static CdmEnvironment getTestEnv() {
+        var testEnv = new CdmEnvironment();
```
sharonluong, line 50: i think here, like in other parts of this PR, it would be good to consistently clarify whether a cdm env variable is a string or a config. i know it's just used here but good to be consistent

sharonluong (changes requested):
Just some clarification of variable names needed and fix a type. Thanks!

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
```diff
@@ -43,11 +42,13 @@ private MigrationProjectFactory() {
      * @param name
      * @param collectionId
      * @param user
+     * @param cdmEnv
      * @return
      * @throws IOException
      */
     public static MigrationProject createMigrationProject(Path path, String name,
-            String collectionId, String user) throws IOException {
+                                                          String collectionId, String user, String cdmEnv)
```
sharonluong, line 50: Is this the ID or the config?

sharonluong (approved):

PR #46, opened 2022-07-25
BXC-3698 - Download unmapped source files

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3698

* Adds a new command (`source_files export_unmapped`) which downloads source files for ever object listed in the source_files.csv which does not have a source path mapped yet. This generally indicates that we are unable to find more recently produced high quality scans in network storage, so we are falling back to the file that was used by CDM.
* Dries up some code in related existing services, and makes adjustments so that they can be integrated with other services more easily (primarily the new `executeDownloadBlock` method)
* Fixes a bug with TestSshServer's password field which was making it so I couldn't test authentication failures.


src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmFileRetrievalService.java
```diff
@@ -50,7 +56,14 @@ public class CdmFileRetrievalService {
      * Download the desc.all file for the collection being migrated
      */
     public void downloadDescAllFile() {
-        downloadFiles(DESC_SUBPATH, getDescAllPath(project));
+        executeDownloadBlock((scpClient) -> {
```
sharonluong, line 59: can you explain the change here?

bbpennel: I refactored `downloadFiles`, which previously contained the behavior in the block just below here, into `executeDownloadBlock`, which now opens up the scp/ssh connection and then executes the provided code block. This is roughly the same as passing code blocks into ruby methods.

The reason for this change was because there are now 3 places where chompb needs to perform scp downloads from CDM. There is a lot in common between the 3, everything that is in `executeDownloadBlock`, but the specifics of what gets downloaded differs for each one (more significantly for the newest one).

sharonluong: i had to re-read this multiple times but i think i get it now. thanks!

sharonluong (approved):

PR #48, opened 2022-08-22
Bxc 3697 sourcefiles status bug

krwong:
[https://jira.lib.unc.edu/browse/BXC-3697](https://jira.lib.unc.edu/browse/BXC-3697)

- modify query and filter for all non-compound objects
- add tests for the unmapped count of group and compound objects

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/status/StatusQueryService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/status/StatusQueryService.java
@@ -66,11 +66,11 @@ public class StatusQueryService {
         }
     }
 
-    // Count non-compound objects, exclude grouped/compound objects
-    private Integer indexedNoncompoundObjectsCountCache;
-    protected int countNoncompoundIndexedObjects() {
-        if (indexedNoncompoundObjectsCountCache != null) {
-            return indexedNoncompoundObjectsCountCache;
+    // Count only file objects, exclude grouped/compound objects
+    private Integer indexedFileObjectsCountCache;
+    protected int countIndexedFileObjects() {
+        if (indexedFileObjectsCountCache != null) {
+            return indexedFileObjectsCountCache;
         }
         CdmIndexService indexService = new CdmIndexService();
         indexService.setProject(project);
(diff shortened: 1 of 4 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
```diff
@@ -249,10 +249,69 @@ public void statusUnmappedVerboseTest() throws Exception {
...
+                "source_files", "status"};
+        executeExpectSuccess(args4);
+
+        assertOutputMatches(".*Unmapped Objects: +3.*");
+    }
+
+    @Test
+    public void statusUnmappedDoesNotContainCompoundObjectsTest() throws Exception {
+        indexCompoundExportSamples();
+        addSourceFile("276_182_E.tif");
```
bbpennel, line 286: for the sake of having a realistic test, we should use filenames that match the collection being used (the filenames listed here are from gilmer). All of the filenames can be seen here:
https://github.com/UNC-Libraries/cdm-to-boxc-migration-util/blob/main/src/test/java/edu/unc/lib/boxc/migration/cdm/services/SipServiceTest.java#L546-L549
We'll also need to specify the correct field to use when calling `source_files generate` below, since it needs to use the `filena` field, like in the example i linked. If we include all these, then the number of unmapped objects should end up as 0, which should clearly illustrate that the compound objects aren't being reported.

src/main/java/edu/unc/lib/boxc/migration/cdm/status/StatusQueryService.java
```diff
@@ -62,6 +66,27 @@ protected int countIndexedObjects() {
...
+    private Integer indexedFileObjectsCountCache;
+    protected int countIndexedFileObjects() {
+        if (indexedFileObjectsCountCache != null) {
+            return indexedFileObjectsCountCache;
+        }
+        CdmIndexService indexService = new CdmIndexService();
+        indexService.setProject(project);
+        try (Connection conn = indexService.openDbConnection()) {
+            Statement stmt = conn.createStatement();
+            // Query for all non-compound objects. If the entry type is null, the object is a individual cdm object
```
bbpennel, line 79: should update the comment here to reflect the "file objects" language

bbpennel (changes requested):

bbpennel (approved):

PR #51, opened 2022-11-08
BXC-3826 - Post Migration Report

bbpennel:
https://jira.lib.unc.edu/browse/BXC-3826
* Running the sip generation command now produces a post_migration_report.csv as a side effect, which lists all the cdm urls and the expected boxc destinations, and object types
* Adds a `verify_migration` command which checks to make sure the boxc urls exist, which is expected to be run after the migration is complete.


src/test/java/edu/unc/lib/boxc/migration/cdm/test/PostMigrationReportTestHelper.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static List<List<String>> parseReport(MigrationProject project) throws Exception {
+        try (
+                Reader reader = Files.newBufferedReader(project.getPostMigrationReportPath());
+                CSVParser csvParser = new CSVParser(reader, PostMigrationReportConstants.CSV_PARSER_FORMAT);
+        ) {
+            var rows = new ArrayList<List<String>>();
+            for (CSVRecord csvRecord : csvParser) {
+                rows.add(StreamSupport.stream(csvRecord.spliterator(), false).collect(Collectors.toList()));
```
bbpennel, line 49: Might be able to switch this to doing csvRecord.toList

src/test/java/edu/unc/lib/boxc/migration/cdm/services/PostMigrationReportServiceTest.java
```diff
@@ -74,7 +74,7 @@ public void addSingleItemTest() throws Exception {
         assertContainsRow(rows, "25",
                 "http://localhost/cdm/singleitem/collection/proj/id/25",
                 "Work",
-                "http://localhost:46888/bxc/record/" + BOXC_ID_1,
+                "http://localhost:46887/bxc/record/" + BOXC_ID_1,
```
sharonluong, line 77: can you make all these `"http://localhost:46887/bxc/record/"` into a constant like `BASE_URL` or something? might be more readable with what's actually different. and if the port changes in the future then it would be easier to update. could use the helper too if you wanted

bbpennel: updated

bbpennel: Updated

sharonluong (approved):

PR #53, opened 2022-11-22
Bxc 3835 new destination mapping command

sharonluong:
This PR adds the ability to add a custom destination mapping by specifying a CDM ID (or comma-delimited list of IDs). It will add a new row to the destination mappings file specifying the CDM ID, the destination mapping and collection (if included).

https://jira.lib.unc.edu/browse/BXC-3835

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/DestinationsCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/DestinationsCommand.java
@@ -133,8 +133,7 @@ public class DestinationsCommand {
         initialize();
         var destinationMappingExists = Files.exists(project.getDestinationMappingsPath());
         if (!destinationMappingExists) {
-          outputLogger.info("FAIL: Destination mapping at path " + project.getDestinationMappingsPath()
-                  + " does not exist");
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/DestinationsCommand.java
```diff
@@ -126,14 +127,37 @@ public int status() throws Exception {
...
 
-    private void validateOptions(GenerateDestinationMappingOptions options) {
+    @Command(name = "add",
+            description = "Add custom destination for individual CDM ID or list of IDs")
+    public int add(@Mixin DestinationMappingOptions options) throws Exception {
+        initialize();
+        var destinationMappingExists = Files.exists(project.getDestinationMappingsPath());
+        if (!destinationMappingExists) {
+          outputLogger.info("FAIL: Destination mapping at path " + project.getDestinationMappingsPath()
+                  + " does not exist");
```
bbpennel, line 137: When using the logger, its preferable to use variable substitution rather than concatenation (for readability, and in some cases it is more efficient)
```suggestion
          outputLogger.info("FAIL: Destination mapping at path {} does not exist", project.getDestinationMappingsPath());
```

bbpennel (approved):
Looks good, one minor suggestion related to the logger

PR #55, opened 2022-12-06
Bxc 3881 postmigration object type report

krwong:
[https://jira.lib.unc.edu/browse/BXC-3881](https://jira.lib.unc.edu/browse/BXC-3881)

- add new `report_migration_types` command to output the counts of new migrated works and files, will only run after post_migration_report.csv is generated

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/MigrationTypeReportCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/MigrationTypeReportCommand.java
@@ -1,0 +1,47 @@
+package edu.unc.lib.boxc.migration.cdm;
+
+import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
+import edu.unc.lib.boxc.migration.cdm.services.MigrationProjectFactory;
+import edu.unc.lib.boxc.migration.cdm.services.MigrationTypeReportService;
+import picocli.CommandLine;
+
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
+
+/**
+ * @author krwong
+ */
+@CommandLine.Command(name = "report_migration_types",
+        description = "Counts new migrated works and files")
+public class MigrationTypeReportCommand implements Callable<Integer> {
+    @CommandLine.ParentCommand
+    private CLIMain parentCommand;
+    private MigrationTypeReportService typeReportService;
+    private MigrationProject project;
+
+
+    public void init() throws Exception {
+        Path currentPath = parentCommand.getWorkingDirectory();
+        project = MigrationProjectFactory.loadMigrationProject(currentPath);
+        typeReportService = new MigrationTypeReportService();
+        typeReportService.setProject(project);
+    }
+
+    @Override
+    public Integer call() throws Exception {
+        try {
+            long start = System.nanoTime();
+            init();
+            String numWorks = String.valueOf(typeReportService.countWorks());
+            String numFiles = String.valueOf(typeReportService.countFiles());
+            outputLogger.info("Number of Works: {}", numWorks);
+            outputLogger.info("Number of Files: {}", numFiles);
+            return 0;
+        } catch (Exception e) {
+            outputLogger.info("Encountered an error while counting new objects", e);
+        }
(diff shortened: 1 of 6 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/MigrationTypeReportCommand.java
```diff
@@ -0,0 +1,47 @@
...
+        typeReportService = new MigrationTypeReportService();
+        typeReportService.setProject(project);
+    }
+
+    @Override
+    public Integer call() throws Exception {
+        try {
+            long start = System.nanoTime();
+            init();
+            String numWorks = String.valueOf(typeReportService.countWorks());
```
bbpennel, line 37: I don't think these need to be case as Strings, the outputLogger should be able to handle longs

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationTypeReportService.java
```diff
@@ -0,0 +1,56 @@
...
+public class MigrationTypeReportService {
+    private static final Logger log = getLogger(MigrationTypeReportService.class);
+
+    private MigrationProject project;
+
+    /**
+     * @return count new works
+     */
+    public long countWorks() throws IOException {
+        var csvParser = openCsvParser();
```
bbpennel, line 27: Should wrap this with a try block so that the parser gets closed after usage. See this example:
https://github.com/UNC-Libraries/cdm-to-boxc-migration-util/blob/main/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PostMigrationReportVerifier.java#L40

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationTypeReportService.java
```diff
@@ -0,0 +1,56 @@
...
+        long numWorks = csvParser.stream()
+                .map(row -> row.get("boxc_obj_type"))
+                .filter(f -> f.toLowerCase().contains("work"))
```
bbpennel, line 30: The object type column gets populated using this constant
`ResourceType.Work.name()`
So it would make sense to use that when performing the counts as well, in case anything gets renamed.

src/test/java/edu/unc/lib/boxc/migration/cdm/MigrationTypeReportCommandIT.java
```diff
@@ -0,0 +1,34 @@
...
+public class MigrationTypeReportCommandIT extends AbstractCommandIT {
+    private MigrationTypeReportService typeReportService;
+
+    @Before
+    public void setup() throws Exception {
+        typeReportService = new MigrationTypeReportService();
+    }
+
+    @Test
+    public void generateMigrationTypeReportTest() throws Exception {
```
bbpennel, line 22: It'd be good to add a test that verifies what happens if there is no post migration report. I'm guessing it'll be a FileNotFound exception currently, which might be fine.

bbpennel (changes requested):
This looks good over all, left a few suggestions

src/main/java/edu/unc/lib/boxc/migration/cdm/MigrationTypeReportCommand.java
```diff
@@ -34,13 +36,14 @@ public Integer call() throws Exception {
...
-            outputLogger.info("Number of Files: {}", numFiles);
+            outputLogger.info("Number of Works: {}", typeReportService.countWorks());
+            outputLogger.info("Number of Files: {}", typeReportService.countFiles());
             return 0;
+        } catch (NoSuchFileException e) {
+            outputLogger.info("Cannot generate migration types report. Post migration report not found: {}",
+                    e.getMessage());
         } catch (Exception e) {
-            outputLogger.info("Encountered an error while counting new objects", e);
+            outputLogger.info("Encountered an error while counting new objects: {}", e.getMessage());
```
bbpennel, line 45: Since I just ran into this while debugging another command, could we include the full stacktrace here? We could send it to the migration.log, which would be `log.error("Encountered an error while counting new objects", e);` (you'll probably have to add a `log` object to this class)

bbpennel (changes requested):

bbpennel (approved):

PR #58, opened 2022-12-21
BXC-3926 adjust query and loop to not depend on file name

sharonluong:
This bugfix makes it so the generateMappings code doesn't need the export Filename if the Blank option is chosen:
https://jira.lib.unc.edu/browse/BXC-3926


bbpennel (approved):
Moving the query construction to a helper method is a definite improvement

PR #59, opened 2023-01-12
Bxc 2878 upgrade to junit 5

krwong:
[https://jira.lib.unc.edu/browse/BXC-2878](https://jira.lib.unc.edu/browse/BXC-2878)

- added junit 5 dependencies to pom.xml
- modified junit-related imports
- changed @ Test annotations for specifying expectations to assertThrows
- changed other annotations (@ Before to @ BeforeEach, @ After to @ AfterEach, etc)
- create temporary folders using @ TempDir instead of @ Rule
- changed wiremock rules to @ WireMockTest
- modified some tests so they work in junit 5
- changed the order of parameters in assertions (output message is last parameter)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,13 +14,14 @@
     <description>Utilities for migrating content from CDM to Box-c</description>
     <properties>
         <picocli.version>4.6.1</picocli.version>
-        <wiremock.version>2.27.2</wiremock.version>
+        <wiremock.version>2.35.0</wiremock.version>
         <sqlite-jdbc.version>3.36.0.1</sqlite-jdbc.version>
         <poi.version>5.1.0</poi.version>
         <log4j-to-slf4j.version>2.17.1</log4j-to-slf4j.version>
         <build-tools.path></build-tools.path>
         <mysql.version>8.0.28</mysql.version>
         <csv.version>1.9.0</csv.version>
+        <junit.jupiter.version>5.9.0</junit.jupiter.version>
     </properties>
 
     <build>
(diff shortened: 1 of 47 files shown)


pom.xml
```diff
@@ -14,13 +14,14 @@
...
         <picocli.version>4.6.1</picocli.version>
-        <wiremock.version>2.27.2</wiremock.version>
+        <wiremock.version>2.35.0</wiremock.version>
         <sqlite-jdbc.version>3.36.0.1</sqlite-jdbc.version>
         <poi.version>5.1.0</poi.version>
         <log4j-to-slf4j.version>2.17.1</log4j-to-slf4j.version>
         <build-tools.path></build-tools.path>
         <mysql.version>8.0.28</mysql.version>
         <csv.version>1.9.0</csv.version>
+        <junit.jupiter.version>5.9.0</junit.jupiter.version>
```
bbpennel, line 24: looks like there is a 5.9.2

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SourceFileServiceTest.java
```diff
@@ -47,13 +48,17 @@ public class SourceFileServiceTest {
...
-                tmpFolder.getRoot().toPath(), PROJECT_NAME, null, "user", CdmEnvironmentHelper.DEFAULT_ENV_ID);
+                tmpFolder, PROJECT_NAME, null, "user", CdmEnvironmentHelper.DEFAULT_ENV_ID);
         Files.createDirectories(project.getExportPath());
 
-        basePath = tmpFolder.newFolder().toPath();
+        File testFolder = new File(String.valueOf(tmpFolder), "testFolder");
+        if (!testFolder.exists()){
+            testFolder.mkdirs();
+        }
+        basePath = testFolder.toPath();
```
bbpennel, line 61: It seems like testFolder shouldn't ever exist at this point, since tmpFolder will have just been created. We should also generally favor java.nio clients over java.io (so Path versus File), unless there's a good reason to use File. So I think we might be able to change this section to:
```suggestion
        basePath = tmpFolder.resolve("testFolder");
        Files.createDirectory(basePath);
```

bbpennel (approved):
This looks great, just a couple minor cleanup suggestions

PR #61, opened 2023-02-06
Bxc 3959 monograph support

sharonluong:
https://jira.lib.unc.edu/browse/BXC-3959

- adjusts for monograph cpd structure differences in assigning children compound objects
- adds tests for exporting and indexing
- adds cpd and desc.all files for test monograph object


bbpennel (approved):
Looks good from my read of it!

PR #70, opened 2023-10-30
BXC-4305 update destinations.csv with archival coll mappings

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4305](https://unclibrary.atlassian.net/browse/BXC-4305)

- `addArchivalCollectionMappings`: method to add archival collection mappings to destination.csv

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
@@ -3,7 +3,10 @@ package edu.unc.lib.boxc.migration.cdm.services;
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
 import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
 import edu.unc.lib.boxc.migration.cdm.options.DestinationMappingOptions;
+import edu.unc.lib.boxc.migration.cdm.validators.DestinationsValidator;
 import edu.unc.lib.boxc.search.api.exceptions.SolrRuntimeException;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
 import org.apache.solr.client.solrj.SolrQuery;
 import org.apache.solr.client.solrj.SolrServerException;
 import org.apache.solr.client.solrj.impl.HttpSolrClient;
@@ -11,6 +14,10 @@ import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.common.SolrDocumentList;
 import org.slf4j.Logger;
 
+import java.io.BufferedWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -96,6 +103,46 @@ public class ArchivalDestinationsService {
             }
         }
         return mapCollNumToPid;
+    }
+
+    /**
+     * Add archival collection mapping(s) to destination mappings file
+     * @param options
+     * @throws Exception
+     */
+    public void addArchivalCollectionMappings(DestinationMappingOptions options) throws Exception {
+        DestinationsValidator.assertValidDestination(options.getDefaultDestination());
+
+        Path destinationMappingsPath = project.getDestinationMappingsPath();
+
+        try (
+                BufferedWriter writer = Files.newBufferedWriter(destinationMappingsPath,
+                        StandardOpenOption.APPEND,
+                        StandardOpenOption.CREATE);
+                CSVPrinter csvPrinter = new CSVPrinter(writer, CSVFormat.Builder.create().build());
+        ) {
+            if (options.getDefaultDestination() != null) {
+                Map<String, String> mapCollNumToPid = generateCollectionNumbersToPidMapping(options);
+                for (Map.Entry<String, String> entry : mapCollNumToPid.entrySet()) {
+                    String collNum = entry.getKey();
+                    String pid = entry.getValue();
+
+                    // destinations.csv columns: id, destination, collection
+                    // if the boxc pid is not null, destination field = the boxc pid and "collection" field is empty
+                    // if the boxc pid is null, destination field is empty and "collection" field = <coll_num>
+                    // the user will need to fill in the "destination" value, which would be a boxc AdminUnit pid
+                    if (pid != null) {
+                        csvPrinter.printRecord(options.getFieldName() + ":" + collNum,
+                                pid,
+                                "");
+                    } else {
+                        csvPrinter.printRecord(options.getFieldName() + ":" + collNum,
+                                "",
+                                collNum);
+                    }
+                }
+            }
+        }
     }
 
     public void setProject(MigrationProject project) {
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
@@ -1,9 +1,9 @@
 package edu.unc.lib.boxc.migration.cdm.services;
 
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
+import edu.unc.lib.boxc.migration.cdm.exceptions.StateAlreadyExistsException;
 import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
 import edu.unc.lib.boxc.migration.cdm.options.DestinationMappingOptions;
-import edu.unc.lib.boxc.migration.cdm.validators.DestinationsValidator;
 import edu.unc.lib.boxc.search.api.exceptions.SolrRuntimeException;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVPrinter;
@@ -15,6 +15,7 @@ import org.apache.solr.common.SolrDocumentList;
 import org.slf4j.Logger;
 
 import java.io.BufferedWriter;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
```diff
@@ -98,6 +105,44 @@ public Map<String, String> generateCollectionNumbersToPidMapping(DestinationMapp
         return mapCollNumToPid;
     }
 
+    /**
+     * Add archival collection mapping(s) to destination mappings file
+     * @param options
+     * @throws Exception
+     */
+    public void addArchivalCollectionMappings(DestinationMappingOptions options) throws Exception {
```
bbpennel, line 113: There doesn't seem to be a test for this method at the moment, it would be good to add some. Some cases we should test would be:

- New file which generates at least one entry with a boxc pid
- New file which generates at least one entry without a boxc pid
- Existing file without force flag (throws an error, to prevent the user from accidentally overwriting whatever is already there)
- Existing file with force flag (overwrites the existing contents with the newly generated mapping)
- No field name is provided

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
```diff
@@ -98,6 +105,44 @@ public Map<String, String> generateCollectionNumbersToPidMapping(DestinationMapp
...
+        Path destinationMappingsPath = project.getDestinationMappingsPath();
+
+        try (
+                BufferedWriter writer = Files.newBufferedWriter(destinationMappingsPath,
+                        StandardOpenOption.APPEND,
+                        StandardOpenOption.CREATE);
+                CSVPrinter csvPrinter = new CSVPrinter(writer, CSVFormat.Builder.create().build());
+        ) {
+            if (options.getFieldName() != null) {
```
bbpennel, line 122: I'd suggest moving this check to the top of the method and doing either an early return or throw an error, since there isn't really a reason to open/create the file if we aren't going to write anything to it, and it'll save having to indent everything an extra time

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
bbpennel: There's also one other bullet point on the ticket that I think needs to be addressed:
> If the destinations.csv file already exists, then overwrite it with a new file if the force flag is provided, otherwise throw an InvalidProjectStateException

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -138,4 +139,129 @@ public void collectionNumberToPidTest() throws Exception {
...
+        testResponse.setResponse(new NamedList<>(Map.of("response", new SolrDocumentList())));
+        when(solrClient.query(solrQueryCaptor.capture())).thenReturn(testResponse);
+    }
+
+    private void solrResponseWithPid() throws Exception {
+        testHelper.indexExportData("mini_keepsakes");
+
+        QueryResponse testResponseA = new QueryResponse();
+        SolrDocument testDocumentA = new SolrDocument();
+        testDocumentA.addField("pid", "40147");
```
bbpennel, line 244: Sorry I didn't catch this early, I'm realizing that the solr document field here doesn't match up with what is in boxc. This class in boxc lists the solr fields:
https://github.com/UNC-Libraries/box-c/blob/main/search-api/src/main/java/edu/unc/lib/boxc/search/api/SearchFieldKey.java#L14
And is available in chompb, so I'd suggest using it as constants in all the cases where we reference a solr field in chompb, both in responses and when generating the query. So something like `SearchFieldKey.ID.getSolrField()` and `SearchFieldKey.COLLECTION_ID.getSolrField()`. It'll help keep things in sync across applications too. We'll need to update `generateCollectionNumbersToPidMapping` as well.

Also, the `id` field would have values like `2720138a-a1ab-4b73-b7bc-6aabe75620c0`, while realistically `40147` is the kind of value we'd find in the `collectionId` field (although in this test it is using contentdm ids, which is fine for testing), so I'd suggest returning a UUID form value here to avoid confusion (I was pretty confused when reading some of the assertions in tests)

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -138,4 +139,129 @@ public void collectionNumberToPidTest() throws Exception {
...
+
+        try (
+                Reader reader = Files.newBufferedReader(destinationMappingsPath);
+                CSVParser csvParser = new CSVParser(reader, CSVFormat.DEFAULT
+                        .withFirstRecordAsHeader()
+                        .withHeader(FieldUrlAssessmentService.URL_CSV_HEADERS)
+                        .withTrim());
+        ) {
+            List<CSVRecord> rows = csvParser.getRecords();
+            assertEquals("", rows.get(0).get(1));
```
bbpennel, line 162: I think the outcome would be a lot clearer if we asserted all the values per row (since there's only 3). We could do something like:
`assertIterableEquals(Arrays.asList("dmrecord:604", "", "604"), rows.get(0));`
i didn't initially understand that the next assertion was expecting a collection id value in the pid column. It'd be good to update the other tests to work this way too

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -138,4 +139,129 @@ public void collectionNumberToPidTest() throws Exception {
...
+
+    @Test
+    public void existingDestinationsFileWithoutForceFlag() throws Exception {
+        solrResponseWithPid();
+        writeCsv(mappingBody("aid:40126,bdbd99af-36a5-4bab-9785-e3a802d3737e,"));
+
+        var options = new DestinationMappingOptions();
+        options.setFieldName("dmrecord");
+
+        try {
```
bbpennel, line 198: We should use the junit 5 form for expecting an exception to be thrown, like this example https://www.baeldung.com/junit-assert-exception#junit-5

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -138,4 +139,129 @@ public void collectionNumberToPidTest() throws Exception {
...
+    public void existingDestinationsFileWithForceFlag() throws Exception {
+        solrResponseWithPid();
+        writeCsv(mappingBody("aid:40126,bdbd99af-36a5-4bab-9785-e3a802d3737e,"));
+
+        var options = new DestinationMappingOptions();
+        options.setFieldName("dmrecord");
+        options.setForce(true);
+
+        service.addArchivalCollectionMappings(options);
+
```
bbpennel, line 217: It'd be good to assert that the file now contains the new mappings, to demonstrate that it overwrote the existing file.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -138,4 +139,129 @@ public void collectionNumberToPidTest() throws Exception {
...
+    }
+
+    @Test
+    public void archivalCollectionMappingsWithoutFieldOption() throws Exception {
+        var options = new DestinationMappingOptions();
+
+        try {
```
bbpennel, line 224: use assertThrows

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
```diff
@@ -43,6 +44,9 @@ public class ArchivalDestinationsService {
     private String solrServerUrl;
     private HttpSolrClient solr;
 
+    public static final String PID = SearchFieldKey.ID.getSolrField();
```
bbpennel, line 47: I'd suggest not naming the constant this, boxc has a PID class that is used all over the place which could conflict or at least caused confusion. Naming it "PID_KEY" would would fine, or "ID_KEY"

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
```diff
@@ -144,7 +148,7 @@ public void addArchivalCollectionMappings(DestinationMappingOptions options) thr
                 }
             }
         } else {
-            throw new Exception("Field option is empty");
+            throw new IllegalArgumentException("Field option is empty");
```
bbpennel, line 151: can `public void addArchivalCollectionMappings(DestinationMappingOptions options) throws Exception {` be changed to something else now? Probably `public void addArchivalCollectionMappings(DestinationMappingOptions options) throws IOException {` at least

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
```diff
@@ -87,7 +91,7 @@ public Map<String, String> generateCollectionNumbersToPidMapping(DestinationMapp
         List<String> listCollectionNumbers = generateCollectionNumbersList(options);
 
         for (String collNum : listCollectionNumbers) {
-            String collNumQuery = "collectionId:" + collNum;
+            String collNumQuery =  COLLECTION_ID + ":" + collNum;
             SolrQuery query = new SolrQuery();
             query.set("q", collNumQuery);
             query.setFilterQueries("resourceType:Collection");
```
bbpennel, line 97: resourceType should also use SearchFieldKey

bbpennel (changes requested):

bbpennel (approved):

PR #71, opened 2023-11-13
BXC-4306 Map SIPs to destinations by archival coll num

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4306](https://unclibrary.atlassian.net/browse/BXC-4306)

- `SipService`: move `initializeDestinations` to end of `initDependencies`, add mapping by archival collection number, add `listCdmIdsByArchivalCollectionId` method, add test
- `ArchivalDestinationsService`: write csv file, add default destination and test
- `CdmToDestMapper`: switch mapping id from DestinationMapping to String
- `SipServiceHelper`: add `generateArchivalDestinationMapping`

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
@@ -75,7 +75,9 @@ public class ArchivalDestinationsService {
                     + CdmIndexService.ENTRY_TYPE_COMPOUND_CHILD + "'" +
                     " OR " + CdmIndexService.ENTRY_TYPE_FIELD + " is null)");
             while (rs.next()) {
-                collectionNumbers.add(rs.getString(1));
+                if (!rs.getString(1).isEmpty()) {
+                    collectionNumbers.add(rs.getString(1));
+                }
             }
             return collectionNumbers;
         } catch (SQLException e) {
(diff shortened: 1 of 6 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsService.java
bbpennel, line 1: while debugging the test, I realized that we will want to skip adding a destination mapping when the mapping field's value is empty, otherwise we'll end up with a "groupa:" entry in the destination mapping which will match everything not in an archival collection, which is really the purpose of the default destination. Searching in solr for "collectionId:" would also return totally random collections.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
```diff
@@ -261,13 +261,52 @@ private void initializeDestinations(SipGenerationOptions options) {
...
+    /**
+     * @param id archival collection id
+     * @return list of CDM ids
+     */
+    private List<String> listCdmIdsByArchivalCollectionId(String id) {
+        String[] splitId = id.split(":");
+        String idField = splitId[0];
+        String idValue = splitId[1];
+
+        List<String> cdmIds = new ArrayList<>();
```
bbpennel, line 286: We should add a check here to return the empty list if the `idValue` is blank, so that we don't end up matching every item that doesn't have an archival collection id (hopefully this shouldn't be able to happen after one of the other comments is addressed, but might as well be safe)

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SipServiceTest.java
```diff
@@ -913,6 +942,99 @@ public void generateSipWithNewCollectionAndRedirectMapping() throws Exception {
...
+        GroupMappingOptions groupOptions = new GroupMappingOptions();
+        groupOptions.setGroupField("groupa");
+        GroupMappingService groupService = testHelper.getGroupMappingService();
+        groupService.generateMapping(groupOptions);
+        groupService.syncMappings(makeDefaultSyncOptions());
+
+        List<MigrationSip> sips = service.generateSips(makeOptions());
+        assertEquals(3, sips.size());
+
+        MigrationSip sip1 = sips.get(0);
```
bbpennel, line 964: My understanding is that each SIP (and therefore destination) should have at least one item mapped to it, so we should have that as our goal here. There should be one SIP for the item going to the default destination, one for group1 and one for group2. 

It would be good to add a comment on each of these sip evaluation blocks to indicate which archival collection the sip belongs to, and probably also add an explicit sip.getDestinationPid assertion so we can directly state that each archival group is going to the right destination pid. So it looks like we should be able to check for DEST_UUID1, DEST_UUID2, and "bdbd99af-36a5-4bab-9785-e3a802d3737e".

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SipServiceTest.java
```diff
@@ -913,6 +942,99 @@ public void generateSipWithNewCollectionAndRedirectMapping() throws Exception {
...
+        DepositDirectoryManager dirManager3 = testHelper.createDepositDirectoryManager(sip3);
+        Model model3 = testHelper.getSipModel(sip3);
+
+        Bag depBag3 = model3.getBag(sip3.getDepositPid().getRepositoryPath());
+        List<RDFNode> depBagChildren3 = depBag3.iterator().toList();
+        assertEquals(4, depBagChildren3.size());
+
+        Resource workResc1 = testHelper.getResourceByCreateTime(depBagChildren3, "2005-11-23");
+        Bag work1Bag = model3.getBag(workResc1);
+        testHelper.assertGroupedWorkPopulatedInSip(workResc1, dirManager3, model3, "grp:groupa:group1", false,
```
bbpennel, line 992: just noting that it won't be groupedWorks anymore after the other changes

bbpennel (changes requested):

bbpennel (approved):

PR #72, opened 2023-11-14
BXC-4336 - Retain newlines in multiline fields

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4336

* Retain newlines in the middle of fields value, so that we can keep line breaks in transcript/full text fields. Whitespace/newlines at the beginning and end of a fields value will still be trimmed off.


krwong (approved):
Looks good!

PR #73, opened 2023-11-28
BXC-4302 archival dest command

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4302](https://unclibrary.atlassian.net/browse/BXC-4302)

- add `map_archival_collections` subcommand and `validateArchivalOptions` to `DestinationsCommand`
- add `WireMockTest`, `assertArchivalCollectionMapping` method, and tests (`generateArchivalDestTest` and `generateArchivalDestNoFieldNameTest`) to `DestinationsCommandIT`


src/main/java/edu/unc/lib/boxc/migration/cdm/DestinationsCommand.java
```diff
@@ -132,6 +138,42 @@ public int add(@Mixin DestinationMappingOptions options) throws Exception {
         }
     }
 
+    @Command(name="map_archival_collections",
+            description = "Generate the destination mappings file by matching CDM field values to " +
+                    "the archival collection number in boxc")
+    public int generateArchival(@Mixin DestinationMappingOptions options) throws Exception {
```
bbpennel, line 144: From a help text perspective, it seems like using the same options class for archival mapping generation and regular mapping generation might be problematic, since regular `generate` doesn't use the fieldName option, and archival doesn't use the cdmId field. The simplest way to address that could be to update the help text for the options to say which command they apply to. Alternatively, we could merge the two commands into one and make "--archival-collections" a flag to trigger that behavior instead. Or we could have two separate options classes, although there is a lot of overlap

src/main/java/edu/unc/lib/boxc/migration/cdm/DestinationsCommand.java
```diff
@@ -48,6 +48,37 @@ public class DestinationsCommand {
     public int generate(@Mixin DestinationMappingOptions options) throws Exception {
         long start = System.nanoTime();
 
+        if (options.isArchivalCollections()) {
```
bbpennel, line 51: just for clarify, i'd suggest moving everything inside this conditional into a private method which gets called here, so here it'd be
```
if (options.isArchivalCollections()) {
    return generateArchivalCollectionMappings(options);
} else {
    return generateBasicMappings(options);
}
```
other something along those lines

src/main/java/edu/unc/lib/boxc/migration/cdm/options/DestinationMappingOptions.java
```diff
@@ -34,10 +34,15 @@ public class DestinationMappingOptions {
                     "The default will be used during SIP generation for objects if they do not have explicit mappings.",})
     private String fieldName;
 
-    @Option(names = { "-f", "--force"},
+    @Option(names = {"-f", "--force"},
             description = "Overwrite destination mapping if one already exists")
     private boolean force;
 
+    @Option(names = {"-ac", "--archival-collections"},
```
bbpennel, line 41: Does picocli support multi-letter short options? I actually don't know, and I'm having a hard time telling from the documentation. I'm fine this if it works though and you've verified it, I guess one of the tests could be updated to reference it with the short form.
https://picocli.info/man/2.x/#_options_and_parameters

bbpennel (approved):

PR #78, opened 2024-01-09
BXC-4380 validate permissions

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4380](https://unclibrary.atlassian.net/browse/BXC-4380)

- `PermissionsCommand`: add `validate` subcommand and tests
- `PermissionsValidator`: add `validateMappings` method, `getPatronRoles` helper method, and tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/validators/PermissionsValidator.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/validators/PermissionsValidator.java
@@ -1,0 +1,118 @@
+package edu.unc.lib.boxc.migration.cdm.validators;
+
+import edu.unc.lib.boxc.auth.api.UserRole;
+import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
+import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
+import edu.unc.lib.boxc.migration.cdm.model.PermissionsInfo;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PermissionsValidator {
+    private MigrationProject project;
+    private List<String> patronRoles;
+
+    /**
+     * Validate the permissions mappings for this project.
+     * @param force if true, incomplete, overlapping and duplicate mappings will be ignored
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/validators/PermissionsValidator.java
```diff
@@ -0,0 +1,118 @@
...
+                String everyone = csvRecord.get(1);
+                String authenticated = csvRecord.get(2);
+
+                // default values
+                if (PermissionsInfo.DEFAULT_ID.equals(id)) {
+                    if (defaultEveryone != null && defaultAuthenticated != null) {
+                        errors.add("Can only map default permissions once, encountered reassignment at line " + i);
+                    } else {
+                        defaultEveryone = everyone;
+                        defaultAuthenticated = authenticated;
```
bbpennel, line 58: It seems like these values are just used for determining if there are multiple default ids. It would be clearer to store a single boolean instead ("hasDefault" or something like that)

src/main/java/edu/unc/lib/boxc/migration/cdm/validators/PermissionsValidator.java
```diff
@@ -0,0 +1,118 @@
...
+public class PermissionsValidator {
+    private MigrationProject project;
+    private List<String> patronRoles;
+
+    /**
+     * Validate the permissions mappings for this project.
+     * @param force if true, incomplete, overlapping and duplicate mappings will be ignored
+     * @return A list of errors. An empty list will be returned for a valid mapping
+     */
+    public List<String> validateMappings(boolean force) {
```
bbpennel, line 28: Is the validator going to be used before sip generation? I'm not sure it makes sense to provide a force flag for just the validate command. Also, i don't think the force flag is doing everything listed in the java doc, it only seems to be used for ignoring blank ids. Maybe it can be removed, or fully implemented if its needed?

src/main/java/edu/unc/lib/boxc/migration/cdm/validators/PermissionsValidator.java
```diff
@@ -0,0 +1,118 @@
...
+                        errors.add("Invalid blank id at line " + i);
+                    }
+                } else if (!PermissionsInfo.DEFAULT_ID.equals(id)) {
+                    errors.add("ID field must be 'default' (for now).");
+                }
+
+                // everyone
+                if (!StringUtils.isBlank(everyone)) {
+                    List<String> patronRoles = getPatronRoles();
+                    if(!patronRoles.contains(everyone)) {
```
bbpennel, line 74: add a space after `if` here and in the authenticated block

bbpennel (changes requested):
This looks good overall, mostly some cleanup and questions about the force flag

bbpennel (approved):

PR #79, opened 2024-01-19
BXC-4407 source_files output: new files, total files mapped, total files in project

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4407](https://unclibrary.atlassian.net/browse/BXC-4407)

- `SourceFilesSummaryService`: `summary`, `newFilesMapped`, `oldFilesMapped`, `totalFilesMapped`, and `totalFilesInProject` methods
- `SourceFilesService`: refactor file paths
- `SourceFilesCommand`, `AccessFilesCommand`, and `AggregateFilesCommand`: initialize `SourceFilesSummaryService`
- add/modify tests in `SourceFilesSummaryServiceTest`, `SourceFilesServiceTest`, and integration tests


src/test/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesSummaryServiceTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Path path1 = testHelper.addSourceFile("25.txt");
+        writeCsv(mappingBody("25,," + path1 +","));
+
+        summaryService.summary(0, Verbosity.NORMAL);
+
+        assertOutputMatches(".*New Files Mapped: +1.*");
+        assertOutputMatches(".*Total Files Mapped: +1.*");
+        assertOutputMatches(".*Total Files in Project: +3.*");
+    }
+
```
bbpennel, line 53: Some extra cases to test for, that the count is correct when:
* there are duplicate entries,
* there are entries with ids not in the index
* nothing is mapped
* there are entries with no path mapped
* everything is mapped (we want to know that it works for more than one mapped file)

bbpennel (changes requested):
Looks reasonable so far. I haven't tried running it yet, but here's some suggestion

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFileService.java
```diff
@@ -154,6 +159,11 @@ public void generateMapping(SourceFileMappingOptions options) throws IOException
             CdmIndexService.closeDbConnection(conn);
         }
 
+        // output summary
+        if (options.getDryRun()) {
+            summaryService.summary(oldNumberFilesMapped, Verbosity.NORMAL);
+        }
```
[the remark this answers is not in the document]
krwong: I checked with Morgan and she does want to do this.

src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
```diff
@@ -117,6 +118,37 @@ public void generateBasicMatchDryRunTest() throws Exception {
...
+        Path tempMappingPath = mappingPath.getParent().resolve("~" + mappingPath.getFileName().toString() + "_new");
+
+        String[] args1 = new String[] {
+                "-w", project.getProjectPath().toString(),
+                "source_files", "generate",
+                "--dry-run",
+                "-b", basePath.toString()};
+        executeExpectSuccess(args1);
+
+        assertTrue(Files.exists(tempMappingPath));
```
bbpennel, line 135: I don't think we want to create an expectation that the temp file is going to exist after a dry run, since we might decide it should be cleaned up later (I almost suggested doing that in an earlier view of this PR). We could instead verify that the regular mappingPath does not exist, since we want to be sure that doesn't show up after a dry run.

This suggestion would apply to all the other dry run tests in this file

src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
```diff
@@ -160,7 +192,7 @@ public void generateUpdateAddSourceFileDryRunTest() throws Exception {
                 "-b", basePath.toString()};
         executeExpectSuccess(args2);
 
-        assertOutputMatches(".*New Files Mapped: +0.*");
+        assertOutputMatches(".*New Files Mapped: +1.*");
         assertOutputMatches(".*Total Files Mapped: +1.*");
```
bbpennel, line 196: shouldn't the total files mapped be 2 since one got mapped in the first generate run, and then another one gets mapped in this dry run?

src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
```diff
@@ -117,6 +118,37 @@ public void generateBasicMatchDryRunTest() throws Exception {
         assertOutputMatches(".*Total Files in Project: +3.*");
```
bbpennel, line 118: I think `generateBasicMatchDryRunTest` and `generateBasicMatchDryRunSummaryTest` are the exact same test now

src/test/java/edu/unc/lib/boxc/migration/cdm/AggregateFilesCommandIT.java
```diff
@@ -68,6 +68,9 @@ public void generateBasicMatchTopTest() throws Exception {
 
         assertTrue(Files.exists(project.getAggregateTopMappingPath()));
         assertFalse(Files.exists(project.getAggregateBottomMappingPath()));
+        assertOutputMatches(".*New Files Mapped: +0.*");
```
bbpennel, line 71: hmm, i don't think these mapping counts are right, it seems like two mappings were supposed to be added? I am guessing what might be the issue is that SourceFilesSummaryService initializes a SourceFileService internally, but it needs to be using an AggregateFileService instead in order to get the correct mapping paths. So SourceFilesSummaryService probably needs to be updated to allow its sourceFileService to be injected so that we can pass in the correct service. This would apply to AccessFiles as well, those counts look off too.

bbpennel (changes requested):

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SourceFileServiceTest.java
```diff
@@ -211,19 +213,34 @@ public void generateMultipleMatchesForSameObjectTest() throws Exception {
...
-        OutputHelper.captureOutput(() -> {
-            testHelper.indexExportData("mini_gilmer");
-            SourceFileMappingOptions options = makeDefaultOptions();
-            options.setDryRun(true);
-            testHelper.addSourceFile("276_182_E.tif");
+    public void generateDryRunVerboseTest() throws Exception {
+        testHelper.indexExportData("mini_gilmer");
+        SourceFileMappingOptions options = makeDefaultOptions();
+        options.setDryRun(true);
+        options.setVerboseOutput(true);
```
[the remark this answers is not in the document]
krwong: Verbose has no impact on SourceFileService. I'll delete the `--verbose-output option` for now, since it isn't added to the summary service yet.

bbpennel (approved):

PR #80, opened 2024-02-08
BXC-4408 sample files output

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4408](https://unclibrary.atlassian.net/browse/BXC-4408)

Display a sample of new files mapped

- `SourceFilesSummaryService`: add `completeListNewFiles`, `sampleListNewFiles`, `listNewFiles` and `showFiles` methods
- modify tests to check for sample of new files mapped

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
@@ -3,12 +3,19 @@ package edu.unc.lib.boxc.migration.cdm.status;
 import edu.unc.lib.boxc.migration.cdm.model.SourceFilesInfo;
 import edu.unc.lib.boxc.migration.cdm.options.Verbosity;
 import edu.unc.lib.boxc.migration.cdm.services.SourceFileService;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 
 import java.io.IOException;
+import java.io.Reader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
@@ -34,12 +41,14 @@ public class SourceFilesSummaryService extends AbstractStatusService {
         int totalFilesMapped = totalFilesMapped();
         int newFilesMapped = newFilesMapped(totalFilesMapped, previousFilesMapped);
         int totalObjects = totalFilesInProject();
+        List<CSVRecord> listFiles = listNewFiles();
 
         if (verbosity.isNormal()) {
             showField("Previous Files Mapped", previousFilesMapped);
             showField("New Files Mapped", newFilesMapped);
             showField("Total Files Mapped", totalFilesMapped);
             showField("Total Files in Project", totalObjects);
+            showFiles(listFiles);
         }
     }
 
@@ -91,6 +100,52 @@ public class SourceFilesSummaryService extends AbstractStatusService {
         return getQueryService().countIndexedFileObjects();
     }
 
+    /**
+     * @return complete list of new files mapped
+     */
+    public List<CSVRecord> completeListNewFiles() {
+        List<CSVRecord> completeListNewFiles = new ArrayList<>();
+
+        try (
+                Reader reader = Files.newBufferedReader(getNewMappingPath());
+                CSVParser csvParser = new CSVParser(reader, CSVFormat.DEFAULT
+                        .withFirstRecordAsHeader()
+                        .withHeader(SourceFilesInfo.CSV_HEADERS)
+                        .withTrim());
+        ) {
+            completeListNewFiles = csvParser.getRecords();
+        } catch (IOException e) {
+            log.error("Failed to list new files", e);
+        }
+        return completeListNewFiles;
+    }
+
+    /**
+     * @return sample list of new files mapped (first and last few rows)
+     */
+    public List<CSVRecord> sampleListNewFiles() {
+        List<CSVRecord> completeListNewFiles = completeListNewFiles();
+        List<CSVRecord> sampleListNewFiles;
+
+        // sample the first 10 and last 10 files mapped
+        if (completeListNewFiles.size() > 20) {
+            sampleListNewFiles = completeListNewFiles.subList(0, 9);
+            sampleListNewFiles.addAll(completeListNewFiles.subList(completeListNewFiles.size() - 10, completeListNewFiles.size()));
+        } else {
+            sampleListNewFiles = completeListNewFiles;
+        }
+
+        return sampleListNewFiles;
+    }
+
+    public List<CSVRecord> listNewFiles() {
+        List<CSVRecord> listNewFiles = new ArrayList<>();
+        if (Files.exists(getNewMappingPath())) {
+            listNewFiles = sampleListNewFiles();
+        }
+        return listNewFiles;
+    }
+
     private Path getNewMappingPath() {
         if (dryRun) {
             return sourceFileService.getTempMappingPath();
@@ -106,6 +161,17 @@ public class SourceFilesSummaryService extends AbstractStatusService {
     public void capturePreviousState() {
         if (Files.exists(getPreviousMappingPath())) {
             setPreviousStateFilesMapped(countFilesMapped(getPreviousMappingPath()));
+        }
+    }
+
+    protected void showFiles(List<CSVRecord> listFiles) {
+        if (listFiles.isEmpty()) {
+            outputLogger.info("Sample unavailable. No new files mapped.");
+        } else {
+            outputLogger.info("{}{}:", INDENT, "Sample of New Files");
+            for (CSVRecord file : listFiles) {
+                outputLogger.info("{}{}* {}", INDENT, INDENT, file);
+            }
         }
     }
 
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
@@ -181,7 +181,7 @@ public class SourceFilesSummaryService extends AbstractStatusService {
 
     private void showFiles(List<CSVRecord> listFiles) {
         if (listFiles.isEmpty()) {
-            outputLogger.info("Sample unavailable. No new files mapped.");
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -20,64 +28,57 @@
  */
 public class SourceFilesSummaryService extends AbstractStatusService {
     private static final Logger log = getLogger(SourceFilesSummaryService.class);
+    private static final String ID_FIELD = "id";
+    private static final String MATCHING_VALUE_FIELD = "matching_value";
+    private static final String SOURCE_FILE_FIELD = "source_file";
+    private static final String POTENTIAL_MATCHES_FIELD = "potential_matches";
```
bbpennel, line 34: We should use the constants in SourceFilesInfo instead of local constants, so that they are more likely to stay in synch

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -91,6 +92,96 @@ public int totalFilesInProject() {
...
+    public List<CSVRecord> loadAllEntries(Path mappingPath) {
+        List<CSVRecord> allEntries = new ArrayList<>();
+
+        try (
+                Reader reader = Files.newBufferedReader(mappingPath);
+                CSVParser csvParser = new CSVParser(reader, CSVFormat.DEFAULT
+                        .withFirstRecordAsHeader()
+                        .withHeader(SourceFilesInfo.CSV_HEADERS)
+                        .withTrim());
+        ) {
```
bbpennel, line 107: I believe you can use `SourceFileService.openMappingsParser()` here to open the parser

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -20,64 +28,57 @@
...
         Set<String> indexedIds = getQueryService().getObjectIdSet();
         Set<String> mappedIds = new HashSet<>();
         try {
-            SourceFilesInfo info = SourceFileService.loadMappings(mappingPath);
-            for (SourceFilesInfo.SourceFileMapping mapping : info.getMappings()) {
-                if (mapping.getSourcePaths() != null && indexedIds.contains(mapping.getCdmId())) {
-                    mappedIds.add(mapping.getCdmId());
+            List<CSVRecord> populatedEntries = loadPopulatedEntries(mappingPath);
+            for (CSVRecord entry : populatedEntries) {
+                if (!entry.get(SOURCE_FILE_FIELD).isEmpty() && indexedIds.contains(entry.get(ID_FIELD))) {
```
bbpennel, line 77: since we're starting from loadPopulatedEntries, we should not need to check `!entry.get(SOURCE_FILE_FIELD).isEmpty()` since that happens inside the other method.

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -20,64 +28,57 @@
...
+                    mappedIds.add(entry.get(ID_FIELD));
                 }
             }
-        } catch (IOException e) {
+        } catch (Exception e) {
             log.error("Failed to load mappings", e);
             outputLogger.info("Failed to load mappings: {}", e.getMessage());
         }
```
bbpennel, line 84: I don't think the try/catch is needed in this method anymore, errors are already being handled in the loadPopulatedEntries method

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -91,6 +92,96 @@ public int totalFilesInProject() {
...
+
+        if (!previousMappings.isEmpty()) {
+            Set<String> previousIdentifiers = getIdentifierSet(previousMappings);
+            Set<String> newIdentifiers = getIdentifierSet(newMappings);
+            newIdentifiers.removeAll(previousIdentifiers);
+            return newMappings.stream().filter(entry -> newIdentifiers.contains(entry.get(ID_FIELD)))
+                    .collect(Collectors.toList());
+        } else {
+            return newMappings;
+        }
```
bbpennel, line 145: I think an early return could simplify the logic here a little bit
```
        List<CSVRecord> newMappings = loadPopulatedEntries(getNewMappingPath());
        if (previousStatePopulatedFiles == null) {
            return newMappings;
        }
        Set<String> previousIdentifiers = getIdentifierSet(previousStatePopulatedFiles);
        Set<String> newIdentifiers = getIdentifierSet(newMappings);
        newIdentifiers.removeAll(previousIdentifiers);
        return newMappings.stream().filter(entry -> newIdentifiers.contains(entry.get(ID_FIELD)))
                .collect(Collectors.toList());
```

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -91,6 +92,96 @@ public int totalFilesInProject() {
         return getQueryService().countIndexedFileObjects();
     }
 
+    /**
+     * @return load list of all CSV records
+     */
+    public List<CSVRecord> loadAllEntries(Path mappingPath) {
```
bbpennel, line 98: We briefly talked about trying to reduce the number of times the CSV file gets loaded, I think at the moment the new mappings file can end up being read 4-5 times per run which could slow things down for large migration projects. One way to address it would be:
1. Add a newStatePopulatedFiles instance variable
2. Introduce getter methods for newStatePopulatedFiles and previousStatePopulatedFiles which set/load their value using loadPopulatedEntries if it hasn't already been set
3. Use the getters in place of loadPopulatedEntries everywhere
4. For `countFilesMapped` we would probably need to change it to receiving a List<CSVRecord> instead of a Path, so that we could pass the correct list in instead of loading it again.

I think we'd be able to adjust `capturePreviousState` to just call `setPreviousStateFilesMapped` again since the loading of entries would be implicit when retrieving the previous state with the getter.

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -91,6 +92,96 @@ public int totalFilesInProject() {
...
+        List<CSVRecord> populatedEntries = loadAllEntries(mappingPath);
+
+        return populatedEntries.stream().filter(entry -> !entry.get(SOURCE_FILE_FIELD).isEmpty())
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * @return list of new CSV records with source file populated
+     */
+    public List<CSVRecord> getListNewEntries() {
```
bbpennel, line 130: We should adjust the name of this method and listNewEntries to make it clearer what each does. We could also adjust the method names to consistently use "Files" or "Entries", since we have a split now after our paired session.

src/test/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryServiceTest.java
```diff
@@ -200,6 +211,52 @@ public void summaryDryRunForceAddSameSourceFile() throws Exception {
...
+        summaryService.setSampleSize(3);
+        summaryService.summary(Verbosity.NORMAL);
+
+        assertOutputMatches(".*Previous Files Mapped: +1.*");
+        assertOutputMatches(".*New Files Mapped: +6.*");
+        assertOutputMatches(".*Total Files Mapped: +7.*");
+        assertOutputMatches(".*Total Files in Project: +161.*");
+        assertOutputContains("26,," + path2);
+        assertOutputContains("28,," + path2);
+        assertOutputContains("30,," + path2);
```
bbpennel, line 237: Might be good to add an assertion that the previous mapping doesn't show up in the output with `assertOutputDoesNotContain`.

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -91,6 +92,96 @@ public int totalFilesInProject() {
...
+    /**
+     * @return sample list of files mapped (if number of mapped files > sample size) or list of all files mapped
+     */
+    public List<CSVRecord> listNewFiles() {
+        List<CSVRecord> listNewFiles = new ArrayList<>();
+        if (Files.exists(getNewMappingPath())) {
+            if (loadPopulatedEntries(getNewMappingPath()).size() > sampleSize) {
+                listNewFiles = sampleListNewFiles();
+            } else {
+                listNewFiles = loadPopulatedEntries(getNewMappingPath());
```
bbpennel, line 179: After looking at some of the tests, i'm realizing this should probably be calling `getListNewEntries`, otherwise the sample will include previously mapped entries when the total number of mapped entries in the new file is less than the sample size. I think we consistently want the listed files to be just the newly added stuff.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -95,68 +85,57 @@ public int totalFilesInProject() {
...
-        List<CSVRecord> newMappings = loadPopulatedEntries(getNewMappingPath());
-
-        if (!previousMappings.isEmpty()) {
-            Set<String> previousIdentifiers = getIdentifierSet(previousMappings);
-            Set<String> newIdentifiers = getIdentifierSet(newMappings);
-            newIdentifiers.removeAll(previousIdentifiers);
-            return newMappings.stream().filter(entry -> newIdentifiers.contains(entry.get(ID_FIELD)))
-                    .collect(Collectors.toList());
-        } else {
+    public List<CSVRecord> getListNewPopulatedFiles() {
```
bbpennel, line 114: how about `listNewlyPopulatedFiles`? My brain has a hard time parsing`getListNew`. And it might also help to update the comment on the method to something like "@return list of CSV records that have been populated since the previous mapping state"

src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -28,15 +25,12 @@
  */
 public class SourceFilesSummaryService extends AbstractStatusService {
```
bbpennel, line 26: One last suggestion, we should try to limit which methods are `public` to just the ones intended to be the public API for this class. I believe that means `summary`, `capturePreviousState`, and the getters/setters (and others if you think they should be accessible elsewhere). It helps make it clearer how the class is intended to be used, and helps avoid invalid/unsafe usages. We can make the others private for now, or protected if they are being used directly by the test or subclasses.

bbpennel (approved):

PR #81, opened 2024-02-14
BXC-4409 source_files verbose output

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4409](https://unclibrary.atlassian.net/browse/BXC-4409)

- `SourceFilesSummaryService`: add verbose output to summary method
- add tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
@@ -38,26 +38,27 @@ public class SourceFilesSummaryService extends AbstractStatusService {
      * @param verbosity
      */
     public void summary(Verbosity verbosity) {
+        if (verbosity.isQuiet()) {
+            return;
+        }
+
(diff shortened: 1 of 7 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/status/SourceFilesSummaryService.java
```diff
@@ -41,14 +41,24 @@ public void summary(Verbosity verbosity) {
         int totalFilesMapped = totalFilesMapped();
         int newFilesMapped = totalFilesMapped - previousStateFilesMapped;
         int totalObjects = totalFilesInProject();
-        List<CSVRecord> listFiles = listNewFiles();
+        List<CSVRecord> listSampleFiles = listNewFiles();
+        List<CSVRecord> listAllFiles = listNewlyPopulatedFiles();
```
bbpennel, line 45: Based off my reading of the ticket, i think we want to display the full contents of the new mapping rather than just the new populated fields, but you could check with morgan to be sure

bbpennel (changes requested):

PR #83, opened 2024-02-19
BXC-4411 add work and file entries to permissions cmd

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4411](https://unclibrary.atlassian.net/browse/BXC-4411)

- `PermissionMappingOptions`: add `--with-works` and `--with-files` flags
- `PermissionsService`: rename `generateDefaultPermissions` method to `generatePermissions`, modify `generatePermissions` to add works/files to csv, add `getIds` helper method
- add tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
@@ -185,11 +185,7 @@ public class PermissionsService {
         // works and files
         if (options.isWithWorks() && options.isWithFiles()) {
             String workAndFileQuery = "select distinct " + CdmFieldInfo.CDM_ID +
-                    " from " + CdmIndexService.TB_NAME
-                    + " where " + CdmIndexService.ENTRY_TYPE_FIELD + " = '" + CdmIndexService.ENTRY_TYPE_GROUPED_WORK + "'"
-                    + " or " + CdmIndexService.ENTRY_TYPE_FIELD + " = '" + CdmIndexService.ENTRY_TYPE_COMPOUND_OBJECT + "'"
-                    + " or " + CdmIndexService.ENTRY_TYPE_FIELD + " = '" + CdmIndexService.ENTRY_TYPE_COMPOUND_CHILD + "'"
-                    + " or " + CdmIndexService.ENTRY_TYPE_FIELD + " is null";
+                    " from " + CdmIndexService.TB_NAME;
             mappedIds = getIds(workAndFileQuery);
         }
 
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
@@ -3,6 +3,7 @@ package edu.unc.lib.boxc.migration.cdm.services;
 import edu.unc.lib.boxc.auth.api.UserRole;
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
 import edu.unc.lib.boxc.migration.cdm.exceptions.StateAlreadyExistsException;
+import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
 import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
 import edu.unc.lib.boxc.migration.cdm.model.PermissionsInfo;
 import edu.unc.lib.boxc.migration.cdm.options.PermissionMappingOptions;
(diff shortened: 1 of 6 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -64,6 +72,54 @@ public void generateDefaultPermissions(PermissionMappingOptions options) throws
                         everyoneField,
                         authenticatedField);
             }
+
+            // works and files
+            if (options.isWithWorks() && options.isWithFiles()) {
```
bbpennel, line 77: I think we could probably dry this up some. One approach could be to have a helper method named `queryForMappingIds(options)` which returns a list of ids (it could return an empty list if no appropriate options are set), and then use that list to add entries to the CSV.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -64,6 +72,54 @@ public void generateDefaultPermissions(PermissionMappingOptions options) throws
...
+                String workAndFileQuery = "select distinct " + CdmFieldInfo.CDM_ID +
+                        " from " + CdmIndexService.TB_NAME
+                        + " where " + ENTRY_TYPE_FIELD + " = '" + CdmIndexService.ENTRY_TYPE_GROUPED_WORK + "'"
```
bbpennel, line 80: For consistency, it'd make sense to always address the constants from other classes either with or without the class name, rather than a mix of both styles.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsServiceTest.java
```diff
@@ -205,6 +205,147 @@ public void generateDefaultPermissionsWithForceFlagTest() throws Exception {
...
+                        .withTrim());
+        ) {
+            List<CSVRecord> rows = csvParser.getRecords();
+            assertIterableEquals(Arrays.asList("default", "canViewMetadata", "canViewMetadata"), rows.get(0));
+            assertIterableEquals(Arrays.asList("25", "canViewMetadata", "canViewMetadata"), rows.get(1));
+            assertIterableEquals(Arrays.asList("26", "canViewMetadata", "canViewMetadata"), rows.get(2));
+            assertIterableEquals(Arrays.asList("27", "canViewMetadata", "canViewMetadata"), rows.get(3));
+        }
+    }
+
```
bbpennel, line 348: We should also test that Files, and Files+Works function correctly when given a project containing compound objects (`mini_keepsakes` for example). Similarly, we should also test Files and Works function correctly with grouped works (you can use `grouped_gilmer` for that)

src/test/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsServiceTest.java
```diff
@@ -205,6 +205,147 @@ public void generateDefaultPermissionsWithForceFlagTest() throws Exception {
...
+        Path permissionsMappingPath = project.getPermissionsPath();
+        var options = new PermissionMappingOptions();
+        options.setWithWorks(true);
+        options.setEveryone(UserRole.canViewMetadata);
+        options.setAuthenticated(UserRole.canViewMetadata);
+
+        service.generatePermissions(options);
+        assertTrue(Files.exists(permissionsMappingPath));
+
+        try (
```
bbpennel, line 220: at this point we should probably make a helper method that reads the mapping file and returns a list of csvrecords, since we're repeating that so many times

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsServiceTest.java
```diff
@@ -193,20 +169,32 @@ public void generateDefaultPermissionsWithForceFlagTest() throws Exception {
...
+        List<CSVRecord> rows = listCsvRecords(permissionsMappingPath);
+        assertIterableEquals(Arrays.asList("default", "canViewMetadata", "canViewMetadata"), rows.get(0));
+        assertIterableEquals(Arrays.asList("25", "canViewMetadata", "canViewMetadata"), rows.get(1));
+        assertIterableEquals(Arrays.asList("26", "canViewMetadata", "canViewMetadata"), rows.get(2));
+        assertIterableEquals(Arrays.asList("27", "canViewMetadata", "canViewMetadata"), rows.get(3));
     }
 
     @Test
-    public void generateWorkPermissionsTest() throws Exception {
+    public void generateWorkPermissionsGroupedWorksTest() throws Exception {
```
bbpennel, line 197: Since grouping doesn't happen automatically you'll need to trigger it similar to how it's happening over in this other test:
https://github.com/UNC-Libraries/cdm-to-boxc-migration-util/blob/main/src/test/java/edu/unc/lib/boxc/migration/cdm/services/AggregateFileMappingServiceTest.java#L211

bbpennel (approved):

PR #84, opened 2024-02-26
BXC-4412 basic set command

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4412](https://unclibrary.atlassian.net/browse/BXC-4412)

- `PermissionsCommand`: add `set` command
- `PermissionMappingOption`: add `--cdm-id` option
- `PermissionsService`: add `setPermissions` method, `updateCsvRecords` helper method, `previousStatePermissions` instance variable
- add tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
@@ -1,6 +1,7 @@
 package edu.unc.lib.boxc.migration.cdm.services;
 
 import edu.unc.lib.boxc.auth.api.UserRole;
+import edu.unc.lib.boxc.migration.cdm.exceptions.InvalidProjectStateException;
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
 import edu.unc.lib.boxc.migration.cdm.exceptions.StateAlreadyExistsException;
 import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
@@ -25,10 +26,10 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
-import static edu.unc.lib.boxc.migration.cdm.services.CdmIndexService.ENTRY_TYPE_COMPOUND_CHILD;
-import static edu.unc.lib.boxc.migration.cdm.services.CdmIndexService.ENTRY_TYPE_FIELD;
+import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
@@ -41,13 +42,14 @@ public class PermissionsService {
     private MigrationProject project;
     private CdmIndexService indexService;
     private List<String> patronRoles;
+    private List<CSVRecord> previousStatePermissions;
 
     /**
      * Generates permission mapping csv
      * @param options permission mapping options
      */
     public void generatePermissions(PermissionMappingOptions options) throws Exception {
-        Path fieldsPath = project.getPermissionsPath();
+        Path permissionsPath = project.getPermissionsPath();
         ensureMappingState(options.isForce());
 
         String everyoneField;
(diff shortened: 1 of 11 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -216,6 +245,61 @@ private List<String> queryForMappedIds(PermissionMappingOptions options) {
         return mappedIds;
     }
 
+    private List<List<String>> updateCsvRecords(PermissionMappingOptions options, Path permissionsMappingPath) throws Exception {
+        List<List<String>> updatedRecords = new ArrayList<>();
+        List<String> cdmIds = new ArrayList<>();
```
bbpennel, line 250: It won't make too big a difference for the scale of records here since it's only checking for one id, but often a `Set`/`HashSet` is a more appropriate data structure in cases where we are collecting ids for lookups or deduplication, since order doesn't matter and `contains` calls are much faster.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -216,6 +245,61 @@ private List<String> queryForMappedIds(PermissionMappingOptions options) {
...
+                            .withFirstRecordAsHeader()
+                            .withHeader(PermissionsInfo.CSV_HEADERS)
+                            .withTrim());
+            ) {
+                permissions = csvParser.getRecords();
+            } catch (IOException e) {
+                log.error("Failed to list permissions", e);
+            }
+
+            setPreviousStatePermissions(permissions);
```
bbpennel, line 298: Are we anticipating needing to hold onto the previous state for comparison in a summary report? Just wondering since as far as i can tell its only being used for one purpose at the moment, which is in `updateCsvRecords`, so I wasn't sure what the caching and prepopulating of the value was for in this case.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsServiceTest.java
```diff
@@ -356,6 +357,82 @@ public void loadPermissionMappingsTest() throws Exception {
...
+        List<CSVRecord> rows = listCsvRecords(permissionsMappingPath);
+        assertIterableEquals(Arrays.asList("default", "canViewMetadata", "canViewMetadata"), rows.get(0));
+        assertIterableEquals(Arrays.asList("25", "none", "none"), rows.get(1));
+        assertIterableEquals(Arrays.asList("26", "none", "none"), rows.get(2));
+        assertIterableEquals(Arrays.asList("27", "canViewMetadata", "canViewMetadata"), rows.get(3));
+    }
+
+    @Test
+    public void setPermissionInvalidIdTest() throws Exception {
+        testHelper.indexExportData("mini_gilmer");
```
bbpennel, line 404: I'd suggest creating an existing CSV file in this test, so that the project is in a valid state for running the set command and won't be sensitive to the order in which validation happens

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
krwong: No, I don't think we need to hold onto the previous state. I was just having trouble getting the previous permissions so I tried using the previous state. I figured out a simpler solution (call `updateCsvRecords` before writing to the permissions csv).

bbpennel (approved):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -248,6 +245,27 @@ private List<String> queryForMappedIds(PermissionMappingOptions options) {
...
+        boolean idExists = false;
+
+        getIndexService();
+        try (Connection conn = indexService.openDbConnection()) {
+            var stmt = conn.prepareStatement(query);
+            stmt.setString(1, id);
+            var rs = stmt.executeQuery();
+            while (rs.next()) {
+                if (!rs.getString(1).isEmpty()) {
+                    idExists = true;
```
bbpennel, line 260: We can return immediately here rather than setting a variable

bbpennel (approved):

PR #87, opened 2024-03-11
BXC-4486 add with-works and with-files to perms set cmd

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4486](https://unclibrary.atlassian.net/browse/BXC-4486)

- `PermissionsService`: modify `updateCsvRecords` to accept `--with-works` and `--with-files` options
- add and modify tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
@@ -102,8 +102,10 @@ public class PermissionsService {
      * @param options permission mapping options
      */
     public void setPermissions(PermissionMappingOptions options) throws Exception {
-        if (!doesIdExistInIndex(options.getCdmId())) {
-            throw new IllegalArgumentException("Id " + options.getCdmId() + " does not exist in this project.");
+        if (options.getCdmId() != null) {
+            if (!doesIdExistInIndex(options.getCdmId())) {
+                throw new IllegalArgumentException("Id " + options.getCdmId() + " does not exist in this project.");
+            }
         }
 
         Path permissionsMappingPath = project.getPermissionsPath();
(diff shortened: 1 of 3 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -295,20 +297,47 @@ private List<List<String>> updateCsvRecords(PermissionMappingOptions options) {
...
+                updatedRecords.add(Arrays.asList(options.getCdmId(), objectType, everyoneField, authenticatedField));
             }
         }
 
-        // add new entry
-        if (!cdmIds.contains(options.getCdmId())) {
-            String objectType = getObjectType(options.getCdmId());
-            updatedRecords.add(Arrays.asList(options.getCdmId(), objectType, everyoneField, authenticatedField));
+        // add or update with-works and with-files entries
+        if (options.isWithWorks() || options.isWithFiles()) {
```
bbpennel, line 319: I haven't confirmed it, but i think if cdmId is set and either one of these options is set, you'll end up with a CSV document with everything twice since the cdmId workflow fills everything in and it all gets added again below. 

I'm wondering if it might make sense to integrate the two workflows by compiling a list of all the ids that need to be updated/added (cdmId + the outcome from addWorkFileRecords), then use the workflow already present for cdmId (loop through the csv, if the id is in the update list then change it otherwise re-add it, and then add all the new stuff at the end). Might be shorter? Alternatively, you could make this block into an `else if` of the cdmId block, in which case the two workflows couldn't happen at the same time. The implementation you currently have seems like it should work well enough otherwise.

bbpennel (changes requested):
Looks good, just one potential issue

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -290,27 +291,59 @@ private String getObjectType(String id) {
...
             }
         }
 
-        // add new entry
-        if (!cdmIds.contains(options.getCdmId())) {
+        // add one new entry
+        if (addedAndUpdatedIds.contains(options.getCdmId())) {
             String objectType = getObjectType(options.getCdmId());
             updatedRecords.add(Arrays.asList(options.getCdmId(), objectType, everyoneField, authenticatedField));
         }
```
bbpennel, line 325: I think we can delete this block and only depend on the loop below to add new entries. To do so, we'd update the block at line 301 to also add an entry to `workAndFileRecords` like `workAndFileRecords.put(options.getCdmId(), getObjectType(options.getCdmId()))`. You might need to change line 305 to add the results from `queryForMappedIds` to `workAndFileRecords` instead of setting the variable to the result, otherwise the cdmId entry would be lost.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PermissionsService.java
```diff
@@ -290,27 +291,59 @@ private String getObjectType(String id) {
...
+        // move default entry to top if it exists
+        List<String> updatedIds = updatedRecords.stream().map(entry -> entry.get(0)).collect(Collectors.toList());
+        int i = 0;
+        for (String id : updatedIds) {
+            if (id.equals(PermissionsInfo.DEFAULT_ID)) {
+                List<String> defaultEntry = updatedRecords.remove(i);
+                updatedRecords.add(0, defaultEntry);
+            }
+            i++;
+        }
```
bbpennel, line 345: I think we could do `updatedIds.indexOf(PermissionsInfo.DEFAULT_ID)` instead of the loop

bbpennel (approved):

PR #88, opened 2024-03-22
BXC-4516 - Do not assign permissions to Files in single file works

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4516

* Only assign file permissions for works that can have more than one file
* Adds assertions to verify single file works don't assign permissions to the files


krwong (approved):
Looks good!

PR #92, opened 2024-05-20
BXC-4512 permit streaming metadata for source file validation

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4512](https://unclibrary.atlassian.net/browse/BXC-4512)

- `SipService` and `SourceFilesValidator`; update validation process in sip generation to consider a file to have a source file mapped to it if streaming metadata fields are present
- `StreamingMetadataService`: switch private method `hasProjectStreamingMetadataField` to public
- add setters/initialize `StreamingMetadataService` in `AggregateFilesCommand`, `SourceFilesCommand`, `StatusCommand`, `ProjectStatusService`, `SourceFilesStatusService`
- fix tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
@@ -221,7 +221,8 @@ public class SipService {
         if (props.getDescriptionsExpandedDate() == null) {
             throw new InvalidProjectStateException("Descriptions must be created and expanded");
         }
-        if (props.getSourceFilesUpdatedDate() == null) {
+        if (props.getSourceFilesUpdatedDate() == null || (props.getSourceFilesUpdatedDate() == null
+                && !streamingMetadataService.hasProjectStreamingMetadataField())) {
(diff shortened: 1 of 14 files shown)


bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
```diff
@@ -221,7 +221,8 @@ private void validateProjectState() {
         if (props.getDescriptionsExpandedDate() == null) {
             throw new InvalidProjectStateException("Descriptions must be created and expanded");
         }
-        if (props.getSourceFilesUpdatedDate() == null) {
+        if (props.getSourceFilesUpdatedDate() == null || (props.getSourceFilesUpdatedDate() == null
+                && !streamingMetadataService.hasProjectStreamingMetadataField())) {
```
bbpennel, line 225: I'm actually not sure we need this exception. Users can generate a blank source file mapping if need be, and it's a little hard to saw what the repercussions further down the pipeline would be for not having a source files mapping at all.

bbpennel (approved):

PR #93, opened 2024-05-21
BXC-4578 add flag to suppress collection level redirects

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4578](https://unclibrary.atlassian.net/browse/BXC-4578)

- `SipGenerationOptions`: add `-sc`/`--suppress-collection-redirect` flag, getter, and setter
- `RedirectMappingService`: add conditional to suppress collection redirects if `options.isSuppressCollectionRedirect` is true
- add options param to `addCollectionRow` method in `RedirectMappingService` and `SipService`
- add test

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
@@ -187,7 +187,7 @@ public class SipService {
                 }
             }
             DisplayProgressUtil.finishProgress();
-            redirectMappingService.addCollectionRow(sips);
+            redirectMappingService.addCollectionRow(options, sips);
             project.getProjectProperties().setSipsGeneratedDate(Instant.now());
             ProjectPropertiesSerialization.write(project);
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/SipService.java
```diff
@@ -187,7 +187,7 @@ public List<MigrationSip> generateSips(SipGenerationOptions options) {
                 }
             }
             DisplayProgressUtil.finishProgress();
-            redirectMappingService.addCollectionRow(sips);
+            redirectMappingService.addCollectionRow(options, sips);
```
bbpennel, line 190: Rather than passing in the SipGenerationOptions and creating a dependency on it there, maybe we should wrap this method call in a `options.isSuppressCollectionRedirect` check instead. In that case we could revert the changes in RedirectMappingService and keep the suppression logic more isolated

bbpennel (changes requested):

bbpennel (changes requested):

bbpennel (approved):

PR #94, opened 2024-05-31
BXC-4596 add sourcefile fields to post_migration_report

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4596](https://unclibrary.atlassian.net/browse/BXC-4596)

- `PostMigrationReportService`: add helper methods `getMatchingValue` and `getSourceFile`, call these methods in `addWorkRow` and `addFileRow`, add setter for `SourceFileService`
- `PostMigrationReportConstants`: add `matching_value` and `source_file` to header, change `VERIFIED_INDEX` to 7
- `SipService`: initialize `SourceFileService`
- fix tests, add `matchingValue` and `sourceFile` to `PostMigrationReportTestHelper` assertions


src/main/java/edu/unc/lib/boxc/migration/cdm/services/PostMigrationReportService.java
```diff
@@ -99,10 +100,19 @@ public void addWorkRow(String cdmObjectId, String boxcWorkId, int childCount, bo
...
         String parentUrl = null;
         String parentTitle = null;
         String objType = ResourceType.Work.name();
-        addRow(cdmObjectId, cdmUrl, objType, boxcUrl, boxcTitle, null, parentUrl, parentTitle, childCount);
+        String sourceFile = null;
+
+        if (isSingleItem) {
+            matchingValue = getMatchingValue(cdmObjectId);
+            sourceFile = getSourceFile(cdmObjectId);
+        }
```
bbpennel, line 112: I'm wondering if we don't really need this since it looks like single items are also being handled down in addFileRow? We probably don't need the same mapping to show up for both the file and the work

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PostMigrationReportService.java
```diff
@@ -175,6 +198,21 @@ private String extractTitle(String cdmId) {
         return null;
     }
 
+    private String getMatchingValue(String cdmId) throws IOException {
+        var sourceFilesInfo = sourceFileService.loadMappings();
+        String matchingValue = sourceFilesInfo.getMappingByCdmId(cdmId).getMatchingValue();
+//        if (matchingValue.isEmpty()) {
```
bbpennel, line 204: can remove commented out code

src/main/java/edu/unc/lib/boxc/migration/cdm/services/PostMigrationReportService.java
```diff
@@ -175,6 +198,21 @@ private String extractTitle(String cdmId) {
         return null;
     }
 
+    private String getMatchingValue(String cdmId) throws IOException {
+        var sourceFilesInfo = sourceFileService.loadMappings();
```
bbpennel, line 202: It doesn't look like there's any sort of caching of the results in sourceFileService, so I would suggest updating PostMigrationReportService to cache the response itself and use that in each of these. Otherwise reading the entire source file mapping twice for file entry could cause some slowness in larger projects.

bbpennel (changes requested):
Looks good over all, just a few suggestions

bbpennel (approved):

PR #95, opened 2024-06-13
BXC-4603 group by multiple fields

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4603](https://unclibrary.atlassian.net/browse/BXC-4603)

- `GroupMappingOptions`: modify`groupField` option to accept a list
- `GroupMappingService`: update `generateMapping` method to use helper methods `generateOneGroupMapping` or `generateMultipleGroupMappings` depending on number of group fields
- add/fix tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
@@ -88,36 +88,80 @@ public class GroupMappingService {
 
             // Return set of all group keys that have at least 2 records in them
             var multiMemberGroupSet = new HashSet<String>();
-            ResultSet groupRs = stmt.executeQuery("select " + options.getGroupField()
-                    + " from " + CdmIndexService.TB_NAME
-                    + " where " + CdmIndexService.ENTRY_TYPE_FIELD + " is null"
-                    + " group by " + options.getGroupField()
-                    + " having count(*) > 1");
-            while (groupRs.next()) {
-                var groupValue = groupRs.getString(1);
-                if (StringUtils.isBlank(groupValue)) {
-                    continue;
-                }
-                multiMemberGroupSet.add(groupValue);
-            }
-
-            ResultSet rs = stmt.executeQuery("select " + CdmFieldInfo.CDM_ID + ", " + options.getGroupField()
-                + " from " + CdmIndexService.TB_NAME
-                + " where " + CdmIndexService.ENTRY_TYPE_FIELD + " is null"
-                + " order by " + CdmFieldInfo.CDM_ID + " ASC");
-            while (rs.next()) {
-                String cdmId = rs.getString(1);
-                String matchedValue = rs.getString(2);
-
-                // Add empty mapping for records either not in groups or in groups with fewer than 2 members
-                if (StringUtils.isBlank(matchedValue) || !multiMemberGroupSet.contains(matchedValue)) {
-                    log.debug("No matching field for object {}", cdmId);
-                    csvPrinter.printRecord(cdmId, null);
-                    continue;
+            if (options.getGroupField().size() == 1) {
+                ResultSet groupRs = stmt.executeQuery("select " + options.getGroupField().get(0)
+                        + " from " + CdmIndexService.TB_NAME
+                        + " where " + CdmIndexService.ENTRY_TYPE_FIELD + " is null"
+                        + " group by " + options.getGroupField().get(0)
+                        + " having count(*) > 1");
+                while (groupRs.next()) {
+                    var groupValue = groupRs.getString(1);
+                    if (StringUtils.isBlank(groupValue)) {
+                        continue;
+                    }
+                    multiMemberGroupSet.add(groupValue);
                 }
 
-                String groupKey = GroupMappingInfo.GROUPED_WORK_PREFIX + options.getGroupField() + ":" + matchedValue;
-                csvPrinter.printRecord(cdmId, groupKey);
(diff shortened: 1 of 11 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -88,50 +88,109 @@ public void generateMapping(GroupMappingOptions options) throws IOException {
...
+                + " where " + CdmIndexService.ENTRY_TYPE_FIELD + " is null"
+                + " group by " + multipleGroups
+                + " having count(*) > 1");
+        while (groupRs.next()) {
+            for (int i = 1; i < numberGroups + 1; i++) {
+                var groupValue = groupRs.getString(i);
                 if (StringUtils.isBlank(groupValue)) {
                     continue;
                 }
                 multiMemberGroupSet.add(groupValue);
```
bbpennel, line 163: I think the value we store in `multiMemberGroupSet` should be the full key of the group, otherwise I think we could end up with unexpected results. Like if we were grouping on `field1` and `field2`, and we have five records:
```
cdmid: 1
field1: a
field2: b

cdmid: 2
field1: a
field2: b

cdmid: 3
field1: c
field2: d

cdmid: 4
field1: c
field2: d

cdmid: 5
field1: a
field2: c
```

I think at the moment we'd end up with a group entry for cdmid 5 since down below it'd return matched fields and both of its values are in `multiMemberGroupSet`.

So I'd suggest joining all the group values (with a "," or "|" delimiter) and adding them as a single key per group in `multiMemberGroupSet`.

I think we also might need to include blank groupValues in the joined key, but we'd want to skip adding to `multiMemberGroupSet` if all the groupValues were blank. Otherwise we could run into key collisions.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -88,50 +88,109 @@ public void generateMapping(GroupMappingOptions options) throws IOException {
...
+                    matchedValues.add(matchedValue);
+                }
+            }
 
                 // Add empty mapping for records either not in groups or in groups with fewer than 2 members
-                if (StringUtils.isBlank(matchedValue) || !multiMemberGroupSet.contains(matchedValue)) {
+                if (matchedValues.isEmpty() || !matchedValues.containsAll(multiMemberGroupSet)) {
                     log.debug("No matching field for object {}", cdmId);
                     csvPrinter.printRecord(cdmId, null);
                     continue;
```
bbpennel, line 185: This block needs to be unindented once

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -177,17 +143,23 @@ private void generateMultipleGroupMapping(GroupMappingOptions options, Statement
                     matchedValues.add(matchedValue);
                 }
             }
+            // Join matched values when grouping by multiple fields
+            String multipleMatchedValues = null;
+            if (numberGroups > 1 && !matchedValues.isEmpty()) {
```
bbpennel, line 148: it's fine to join even when there is only field being grouped by, you'll just get back the one value. Even joining on an empty list is often fine, you'll just get an empty string. If we always set a multipleMatchedValues value then you could use that for all the conditions in the `if` down at line 153 instead of using matchedValues some of the time, as well as simplify the number of cases.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -109,58 +105,28 @@ public void generateMapping(GroupMappingOptions options) throws Exception {
...
+            List<String> groupValues = new ArrayList<>();
             for (int i = 1; i < numberGroups + 1; i++) {
                 var groupValue = groupRs.getString(i);
                 if (StringUtils.isBlank(groupValue)) {
                     continue;
                 }
-                multiMemberGroupSet.add(groupValue);
+                groupValues.add(groupValue);
+            }
+            if (!groupValues.isEmpty()) {
```
bbpennel, line 127: This is good, but one case we're still not covering is key collisions with blank fields. For example, currently these two records would produce the same key, even though their data is different:
```
cdmid: 1
field1: a
field2:

cdmid: 2
field1:
field2: a
```
They'd both end up with "a" as the key. If we leave blanks in the key, we'd get "a," and ",a" as the keys, so they wouldn't get grouped together incorrectly. 

So I'd suggest allowing blanks into `groupValues` and `matchedValues`, and then adding a helper method for building the key which take a list of values and would:
1. Return null or empty string if every value in the list is empty
2. Otherwise, join all the values like you're currently doing

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/GroupMappingService.java
```diff
@@ -197,6 +185,14 @@ private void ensureMappingState(GroupMappingOptions options) {
         }
     }
 
+    private String buildKey(List<String> values) {
+        String joinedValues = null;
+        if (!values.isEmpty()) {
```
bbpennel, line 190: I don't think this will catch if all the entries in the list are empty, just if the list itself is empty. It's possible that it could result in entries with all blanks for the group field would produce the key ",," (for a 3 field grouping request), which might result in everything with blank fields getting grouped together (although I would have thought the tests would have caught that?).  One way to check if they are all empty is `values.stream().allMatch(String::isEmpty)`

bbpennel (approved):

PR #98, opened 2024-07-15
BXC-4635 init project from filesystem

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4635](https://unclibrary.atlassian.net/browse/BXC-4635)

- `InitializeProjectCommand`: add `-s, --source` option to init a new chompb project that doesn't need to be linked to a CDM collection
- `MigrationProjectFactory`: `CdmCollectionId` is `null` when `CdmEnvId` is `null`
- add test

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
@@ -54,8 +54,11 @@ public class InitializeProjectCommand implements Callable<Integer> {
                 "If the project name is different from the CDM collection ID, then use -c to specify the ID." })
     private String projectName;
     @Option(names = {"-s", "--source"},
-            description = {"Initialize a new project that doesn't need to be linked with a CDM collection."})
-    private String newProject;
+            description = {"Specify the source of migration data when initializing a new project. Accepted values: " +
+                    "cdm (CDM collection), files (filesystem, doesn't need to be linked with CDM collection)." +
+                    "Defaults to cdm."},
+            defaultValue = "cdm")
+    private String projectSource;
 
     private CdmFieldService fieldService;
     private CloseableHttpClient httpClient;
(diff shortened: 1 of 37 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
```diff
@@ -53,6 +53,9 @@ public class InitializeProjectCommand implements Callable<Integer> {
                 "If not specified, then the project will be initialized in the current directory.",
                 "If the project name is different from the CDM collection ID, then use -c to specify the ID." })
     private String projectName;
+    @Option(names = {"-s", "--source"},
+            description = {"Initialize a new project that doesn't need to be linked with a CDM collection."})
+    private String newProject;
```
bbpennel, line 58: The idea with this parameter is that it specifies the source of the data being migrated. So for the time being it would take values like "files" and "cdm", and later it might take "aspace". So I think we should adjust the description and the variable name accordingly, maybe projectSource or something with source in it. And it would default to "cdm". We should also validate its value.

src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
```diff
@@ -87,48 +90,60 @@ public Integer call() throws Exception {
...
-            fieldInfo = fieldService.retrieveFieldsForCollection(collId);
-        } catch (IOException | MigrationException e) {
-            log.error("Failed to retrieve field information for collection in project", e);
-            outputLogger.info("Failed to retrieve field information for collection {} in project {}:\n{}",
-                    collId, projDisplayName, e.getMessage());
-            return 1;
-        }
+        if (newProject != null && !newProject.isBlank()) {
+            // Instantiate the project
+            try {
```
bbpennel, line 101: Might help with readability to have separate `initCdmProject` and `initFilesProject` private methods (which returns an Integer)

src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
```diff
@@ -87,48 +90,60 @@ public Integer call() throws Exception {
...
+                MigrationProjectFactory.createMigrationProject(
```
bbpennel, line 102: It would probably also make sense to store the value of this property into the project properties file, in case we need to check the value in later steps to see if the user is allowed to do them, or if chompb needs to pull data from a particularly location, etc. We could have separate `MigrationProjectFactory.createCdmMigrationProject` and `MigrationProjectFactory.createFilesMigrationProject` (they could both call MigrationProjectFactory.createMigrationProject internally) which would allow more control over what params get passed in. Or just keep one method and pass the attribute in as another param.

src/test/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommandIT.java
```diff
@@ -186,6 +184,23 @@ public void initNoEnvMappingPathTest() throws Exception {
...
+                "-w", baseDir.toString(),
+                "--env-config", chompbConfigPath,
+                "init",
+                "-p", "test_file_project",
+                "-s", "files" };
+        executeExpectSuccess(initArgs);
+
+        MigrationProject project = MigrationProjectFactory.loadMigrationProject(baseDir.resolve("test_file_project"));
+        MigrationProjectProperties properties = project.getProjectProperties();
+        assertNewProjectPropertiesSet(properties, "test_file_project", null);
```
bbpennel, line 199: We should probably just fold this into the test since this is the only place it is used. Otherwise, we should update the name of the method to something like `assertFilesProjectPropertiesSet`

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
```diff
@@ -21,10 +21,18 @@ public class MigrationProjectFactory {
...
+    public static MigrationProject createCdmMigrationProject(Path path, String name,
+                                                             String collectionId, String user,
+                                                             String cdmEnvId, String bxcEnvId)
             throws IOException {
-        return createMigrationProject(path, name, collectionId, user, cdmEnvId, null);
+        return createMigrationProject(path, name, collectionId, user, cdmEnvId, bxcEnvId, "cdm");
+    }
+
+    public static MigrationProject createFilesMigrationProject(Path path, String name,
+                                                          String collectionId, String user,
```
bbpennel, line 32: We can remove collectionId here since that only applies to CDM projects, and pass null into createMigrationProject

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
```diff
@@ -21,10 +21,18 @@ public class MigrationProjectFactory {
...
+                                                               String bxcEnvId, String projectSource)
```
bbpennel, line 33: Can also leave off projectSource and pass a constant into createMigrationProject like in createCdmMigrationProject

src/main/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommand.java
```diff
@@ -87,11 +93,24 @@ public Integer call() throws Exception {
...
 
         Path currentPath = parentCommand.getWorkingDirectory();
         String projDisplayName = projectName == null ? currentPath.getFileName().toString() : projectName;
+        Integer integer = null;
+
+        if (projectSource.equalsIgnoreCase("cdm")) {
+            integer = initCdmProject(config, currentPath, projDisplayName, start);
+        } else if (projectSource.equalsIgnoreCase("files")) {
+            integer = initFilesProject(currentPath, projDisplayName, start);
+        }
```
bbpennel, line 105: We can add an `else` and have it log an error message about it being an invalid value (and probably set integer to -1, i'm not really sure what returning a null is going to do).

src/main/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactory.java
```diff
@@ -21,10 +21,18 @@ public class MigrationProjectFactory {
...
     }
 
-    public static MigrationProject createMigrationProject(Path path, String name,
-                                                          String collectionId, String user, String cdmEnvId)
+    public static MigrationProject createCdmMigrationProject(Path path, String name,
+                                                             String collectionId, String user,
+                                                             String cdmEnvId, String bxcEnvId)
             throws IOException {
-        return createMigrationProject(path, name, collectionId, user, cdmEnvId, null);
+        return createMigrationProject(path, name, collectionId, user, cdmEnvId, bxcEnvId, "cdm");
```
bbpennel, line 28: I think we should make constants like PROJECT_SOURCE_CDM for "cdm" and "files", and use those everywhere to try to ensure consistency. I'm trying to think of the best place to put them. It might make sense in either MigrationProject or MigrationProjectProperties, or in a new constants class of some sort.

src/test/java/edu/unc/lib/boxc/migration/cdm/InitializeProjectCommandIT.java
```diff
@@ -186,6 +184,29 @@ public void initNoEnvMappingPathTest() throws Exception {
...
+                "init",
+                "-p", "test_file_project",
+                "-s", "files" };
+        executeExpectSuccess(initArgs);
+
+        MigrationProject project = MigrationProjectFactory.loadMigrationProject(baseDir.resolve("test_file_project"));
+        MigrationProjectProperties properties = project.getProjectProperties();
+        assertEquals(USERNAME, properties.getCreator());
+        assertEquals("test_file_project", properties.getName(), "Project name did not match expected value");
+        assertNull(properties.getCdmCollectionId(), "CDM Collection ID did not match expected value");
```
bbpennel, line 201: Can probably remove this message, I think it was copied from an assertEquals

src/test/java/edu/unc/lib/boxc/migration/cdm/services/ArchivalDestinationsServiceTest.java
```diff
@@ -47,6 +47,7 @@
 
 public class ArchivalDestinationsServiceTest {
     private static final String PROJECT_NAME = "proj";
+    protected final static String PROJECT_SOURCE = "cdm";
```
bbpennel, line 50: Hopefully we can replace all these one off constants in tests with the central one suggested earlier

src/test/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactoryTest.java
bbpennel, line 1: We should add a new test method to this file to directly verify that createFilesMigrationProject works.

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/MigrationTypeReportServiceTest.java
```diff
@@ -38,8 +38,8 @@ public class MigrationTypeReportServiceTest {
     @BeforeEach
     public void setup() throws Exception {
         project = MigrationProjectFactory.createMigrationProject(
-                tmpFolder, "proj", null, "user",
-                CdmEnvironmentHelper.DEFAULT_ENV_ID, BxcEnvironmentHelper.DEFAULT_ENV_ID, PROJECT_SOURCE);
+                tmpFolder, "proj", null, "user", CdmEnvironmentHelper.DEFAULT_ENV_ID,
+                BxcEnvironmentHelper.DEFAULT_ENV_ID, MigrationProject.PROJECT_SOURCE_CDM);
```
bbpennel, line 42: Delete the old PROJECT_SOURCE constant

src/test/java/edu/unc/lib/boxc/migration/cdm/services/RedirectMappingIndexServiceTest.java
```diff
@@ -49,8 +49,8 @@ public class RedirectMappingIndexServiceTest {
     @BeforeEach
     public void setup() throws Exception {
         project = MigrationProjectFactory.createMigrationProject(
-                tmpFolder, PROJECT_NAME, null, "user",
-                CdmEnvironmentHelper.DEFAULT_ENV_ID, BxcEnvironmentHelper.DEFAULT_ENV_ID, PROJECT_SOURCE);
+                tmpFolder, PROJECT_NAME, null, "user", CdmEnvironmentHelper.DEFAULT_ENV_ID,
+                BxcEnvironmentHelper.DEFAULT_ENV_ID, MigrationProject.PROJECT_SOURCE_CDM);
```
bbpennel, line 53: Delete the old PROJECT_SOURCE constant

src/test/java/edu/unc/lib/boxc/migration/cdm/services/MigrationProjectFactoryTest.java
```diff
@@ -202,6 +203,38 @@ public void loadValidProject() throws Exception {
...
+        assertNull(returnedProperties.getCdmEnvironmentId());
+
+        Path propertiesPath = projectLoaded.getProjectPropertiesPath();
+        assertTrue(Files.exists(propertiesPath), "Properties object must exist");
+        MigrationProjectProperties propertiesFile = ProjectPropertiesSerialization.read(propertiesPath);
+        assertEquals(USERNAME, propertiesFile.getCreator());
+        assertEquals(PROJ_NAME, propertiesFile.getName(), "Project name did not match expected value");
+        assertNull(propertiesFile.getCdmCollectionId());
+        assertNotNull(propertiesFile.getCreatedDate(), "Created date not set");
+        assertNull(propertiesFile.getCdmEnvironmentId());
```
bbpennel, line 231: We should also verify that `getProjectSource` returns "files"

bbpennel (changes requested):

bbpennel (approved):

PR #104, opened 2024-08-09
BXC-4652 implement index --from-csv

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4652](https://unclibrary.atlassian.net/browse/BXC-4652)

- `CdmIndexOptions`: add `--from-csv` option
- `CdmIndexCommand`: initialize `ExportObjectsService` and add `CdmIndexOptions`
- `CdmFieldService`: add methods to retrieve, validate, and load fields from `exported_objects.csv`
- `CdmIndexService`: add options param to `createDatabase` method, add `indexAllFromCsv` method
- `ExportObjectsService`: add methods `loadExportObjects`, `recordToExportFile`, and `openMappingsParser`
- add tests and test resources

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
@@ -7,6 +7,8 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.concurrent.Callable;
 
+import edu.unc.lib.boxc.migration.cdm.options.CdmIndexOptions;
+import edu.unc.lib.boxc.migration.cdm.services.ExportObjectsService;
 import org.slf4j.Logger;
 
 import edu.unc.lib.boxc.migration.cdm.exceptions.StateAlreadyExistsException;
@@ -14,6 +16,7 @@ import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
 import edu.unc.lib.boxc.migration.cdm.services.CdmFieldService;
 import edu.unc.lib.boxc.migration.cdm.services.CdmIndexService;
 import edu.unc.lib.boxc.migration.cdm.services.MigrationProjectFactory;
+import picocli.CommandLine.Mixin;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
 import picocli.CommandLine.ParentCommand;
@@ -22,7 +25,8 @@ import picocli.CommandLine.ParentCommand;
  * @author bbpennel
  */
 @Command(name = "index",
-        description = "Index the exported CDM records for this project. Must be run after a complete export.")
+        description = "Index the exported CDM records for this project. Must be run after a complete export " +
+                "unless indexing from exported_objects.csv (exporting form the filesystem).")
 public class CdmIndexCommand implements Callable<Integer> {
     private static final Logger log = getLogger(CdmIndexCommand.class);
     @ParentCommand
@@ -34,7 +38,11 @@ public class CdmIndexCommand implements Callable<Integer> {
 
     private CdmFieldService fieldService;
     private CdmIndexService indexService;
+    private ExportObjectsService exportObjectsService;
     private MigrationProject project;
+
+    @Mixin
+    private CdmIndexOptions options;
 
     @Override
     public Integer call() throws Exception {
@@ -43,8 +51,8 @@ public class CdmIndexCommand implements Callable<Integer> {
         try {
             initialize();
 
-            indexService.createDatabase(force);
-            indexService.indexAll();
+            indexService.createDatabase(force, options);
+            indexService.index(options);
             // Display any warning messages to user
             if (!indexService.getIndexingWarnings().isEmpty()) {
                 indexService.getIndexingWarnings().forEach(msg -> outputLogger.info(msg));
@@ -67,8 +75,11 @@ public class CdmIndexCommand implements Callable<Integer> {
         Path currentPath = parentCommand.getWorkingDirectory();
         project = MigrationProjectFactory.loadMigrationProject(currentPath);
         fieldService = new CdmFieldService();
+        exportObjectsService = new ExportObjectsService();
+        exportObjectsService.setProject(project);
         indexService = new CdmIndexService();
         indexService.setFieldService(fieldService);
+        indexService.setExportObjectsService(exportObjectsService);
         indexService.setProject(project);
     }
 }
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
@@ -4,9 +4,12 @@ import static edu.unc.lib.boxc.migration.cdm.util.CLIConstants.outputLogger;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.Callable;
 
+import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
+import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
 import edu.unc.lib.boxc.migration.cdm.options.CdmIndexOptions;
 import edu.unc.lib.boxc.migration.cdm.services.ExportObjectsService;
 import org.slf4j.Logger;
@@ -25,8 +28,8 @@ import picocli.CommandLine.ParentCommand;
  * @author bbpennel
  */
 @Command(name = "index",
-        description = "Index the exported CDM records for this project. Must be run after a complete export " +
-                "unless indexing from exported_objects.csv (exporting form the filesystem).")
+        description = "Populate the index of object records for this project. Must be run after " +
+                "exporting source metadata or providing a CSV file.")
 public class CdmIndexCommand implements Callable<Integer> {
     private static final Logger log = getLogger(CdmIndexCommand.class);
     @ParentCommand
@@ -50,6 +53,16 @@ public class CdmIndexCommand implements Callable<Integer> {
 
         try {
             initialize();
+
+            // if user provides csv, check that it exists
+            if (options.getCsvFile() != null) {
+                if (Files.exists(options.getCsvFile())) {
+                    CdmFieldInfo csvExportFields = fieldService.retrieveFieldsFromCsv(options.getCsvFile());
+                    fieldService.persistCsvFieldsToProject(project, csvExportFields);
+                } else {
+                    throw new MigrationException("No csv file exists in " + options.getCsvFile());
+                }
+            }
 
             indexService.createDatabase(force, options);
             indexService.index(options);
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
@@ -11,7 +11,6 @@ import java.util.concurrent.Callable;
 import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
 import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
 import edu.unc.lib.boxc.migration.cdm.options.CdmIndexOptions;
-import edu.unc.lib.boxc.migration.cdm.services.ExportObjectsService;
 import org.slf4j.Logger;
 
 import edu.unc.lib.boxc.migration.cdm.exceptions.StateAlreadyExistsException;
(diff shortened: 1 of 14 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommandIT.java
```diff
@@ -121,6 +121,26 @@ public void indexWithWarningsTest() throws Exception {
         assertOutputContains("CPD file referenced by object 604 in desc.all was not found");
     }
 
+    @Test
+    public void indexFromCsvTest() throws Exception {
+        initProject();
+        Files.createDirectories(project.getExportPath());
+
+        Files.copy(Paths.get("src/test/resources/files/cdm_export_fields.csv"), project.getFieldsPath());
```
bbpennel, line 129: The `cdm_fields.csv` should not exist before calling the indexing command, it should be getting created as a side effect of the indexing.

We should assert that `cdm_fields.csv` exists after running the command though.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -335,6 +355,43 @@ private String indexFieldType(String exportField) {
...
+    public void indexAllFromCsv(CdmIndexOptions options) throws IOException {
+        assertObjectsExported();
+
+        CdmFieldInfo fieldInfo = fieldService.loadFromCsvFieldsFromProject(project);
+        List<String> exportFields = fieldInfo.listAllExportFields();
+        recordInsertSqlTemplate = makeInsertTemplate(exportFields);
+
+        try (
+                var conn = openDbConnection();
+                var csvParser = ExportObjectsService.openMappingsParser(Path.of(options.getCsvFile()));
```
bbpennel, line 371: The CSV file is not going to be exactly like the export file format. See my comment below about the number of columns. So I think we just want to parse the provided CSV file as a generic CSV file rather than as an exported_objects.csv. The purpose of having the user export and then import a CSV file is so that they can add more info per record than we get from just the file system (which at the moment is basically just file paths).

If we need to provide a list of header fields, that would probably be `exportFields`.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ExportObjectsService.java
```diff
@@ -56,6 +59,53 @@ private void validateProjectState() {
         }
     }
 
+    /**
+     * @return the export objects mapping info for the configured project
+     * @throws IOException
+     */
+    public ExportObjectsInfo loadExportObjects() throws IOException {
```
bbpennel, line 66: I'm guessing we won't need the changes to this file

src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmFieldServiceTest.java
```diff
@@ -342,6 +345,57 @@ public void retrieveValidateAndReloadRoundTripTest() throws Exception {
...
+        assertHasFieldWithValue(ExportObjectsInfo.FILE_PATH, ExportObjectsInfo.FILE_PATH, ExportObjectsInfo.FILE_PATH,
+                false, null, null, null,
+                null, null, fields);
+        assertHasFieldWithValue(ExportObjectsInfo.FILENAME, ExportObjectsInfo.FILENAME, ExportObjectsInfo.FILENAME,
+                false, null, null, null,
+                null, null, fields);
+    }
+
+    @Test
+    public void persistAndLoadFromCsvFieldsTest() throws Exception {
```
bbpennel, line 368: Might not need this test

src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
```diff
@@ -629,6 +646,48 @@ public void buildDocumentWithLongFieldTest() throws Exception {
         assertEquals("7", rootEl.getChildText("dmrecord"));
     }
 
+    @Test
+    public void indexFromCsvTest() throws Exception {
```
bbpennel, line 650: we should probably add another test to verify that this works for a CSV file with more fields than the default ones from the export. I'd suggest using the fieldService.retrieveFieldsFromCsv method to get the fields from the provided CSV and then persisting them in this test and the new one as well, rather than using `cdm_export_fields.csv`

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmFieldService.java
```diff
@@ -234,6 +235,101 @@ private void validateFieldName(String field, String headerField, int line, Set<S
...
+        CdmFieldInfo fieldInfo = new CdmFieldInfo();
+
+        try (
+            Reader reader = Files.newBufferedReader(exportedObjectsPath);
+            CSVParser parser = new CSVParser(reader, CSVFormat.DEFAULT
+                    .withHeader(ExportObjectsInfo.CSV_HEADERS)
+                    .withTrim());
+        ) {
+            List<String> headers = parser.getHeaderNames();
+            for (String header : headers) {
```
bbpennel, line 253: Since the CDM_ID field is required all over the place, I think we should translate the incoming record_id field into the dmrecord field. Or maybe we should have had the export command put the field in as dmrecord to begin with, which we could change in this PR. I think we are going to want to rename the field from dmrecord into record_id everywhere in the future, but we don't need to do that in this PR.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
```diff
@@ -34,17 +41,31 @@ public class CdmIndexCommand implements Callable<Integer> {
...
+            if (options.getCsvFile() != null) {
+                if (Files.exists(options.getCsvFile())) {
+                    CdmFieldInfo csvExportFields = fieldService.retrieveFieldsFromCsv(options.getCsvFile());
+                    fieldService.persistCsvFieldsToProject(project, csvExportFields);
+                } else {
+                    throw new MigrationException("No csv file exists in " + options.getCsvFile());
+                }
+            }
+
+            indexService.createDatabase(force, options);
```
bbpennel, line 67: just noticed we still have the `force` flag as a separate option, but we have CdmIndexOptions now. We should move it from this class into the options, and then createDatabase only needs one param

src/main/java/edu/unc/lib/boxc/migration/cdm/CdmIndexCommand.java
```diff
@@ -67,8 +88,11 @@ private void initialize() throws IOException {
         Path currentPath = parentCommand.getWorkingDirectory();
         project = MigrationProjectFactory.loadMigrationProject(currentPath);
         fieldService = new CdmFieldService();
+        exportObjectsService = new ExportObjectsService();
```
bbpennel, line 91: I don't think we need ExportObjectsService in CdmIndexService anymore, so we should remove it from  anywhere it was being initialized for CdmIndexService

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmFieldService.java
```diff
@@ -234,6 +234,85 @@ private void validateFieldName(String field, String headerField, int line, Set<S
...
+        return fieldInfo;
+    }
+
+    /**
+     * Persist the field information out to the project
+     * @param project
+     * @param fieldInfo
+     * @throws IOException
+     */
+    public void persistCsvFieldsToProject(MigrationProject project, CdmFieldInfo fieldInfo) throws IOException {
```
bbpennel, line 271: I could be wrong, but I feel like we don't need this method and can use `persistFieldsToProject` in place of it? The field info should be populated the same whether it came from CDM or a CSV. I know CDM has way more attributes than we need, but it doesn't do any harm for them to be present and set null/blank.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -295,12 +313,19 @@ private void indexObject(Connection conn, List<String> exportFieldValues)
      * @param force
      * @throws IOException
      */
-    public void createDatabase(boolean force) throws IOException {
+    public void createDatabase(boolean force, CdmIndexOptions options) throws IOException {
         ensureDatabaseState(force);
+        List<String> exportFields;
+
+        if (options.getCsvFile() != null) {
```
bbpennel, line 320: I think it's okay to have the same behavior for CSV or CDM. CSV fields should be loaded with `loadFieldsFromProject` in either case, and I think `exportFields.addAll(MIGRATION_FIELDS);` is fine in both cases too.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -335,6 +360,52 @@ private String indexFieldType(String exportField) {
...
 
+    /**
+     * Indexes all exported objects for this project
+     * @param options
+     * @throws IOException
+     */
+    public void indexAllFromCsv(CdmIndexOptions options) throws IOException {
+        assertObjectsExported(options);
+
+        CdmFieldInfo fieldInfo = fieldService.loadFromCsvFieldsFromProject(project);
```
bbpennel, line 371: use `loadFieldsFromProject`. The only time we should need to load the fields from the CSV file is when populating the field info object for persisting purposes

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -354,6 +425,10 @@ public void setFieldService(CdmFieldService fieldService) {
         this.fieldService = fieldService;
     }
 
+    public void setExportObjectsService(ExportObjectsService exportObjectsService) {
```
bbpennel, line 428: Can remove this

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -335,6 +360,52 @@ private String indexFieldType(String exportField) {
...
+                        .withFirstRecordAsHeader()
+                        .withHeader(String.valueOf(exportFields))
+                        .withTrim());
+        ) {
+            for (CSVRecord csvRecord : csvParser) {
+                if (!csvRecord.get(0).isEmpty()) {
+                    List<String> fieldValues = new ArrayList<>();
+                    for (int i = 0; i < csvRecord.size(); i++) {
+                        fieldValues.add(csvRecord.get(i));
+                    }
```
bbpennel, line 389: It looks like there is `csvRecord.toList()` which will do the same thing as this.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -335,6 +360,52 @@ private String indexFieldType(String exportField) {
...
+            throw new MigrationException("Failed to read export files", e);
+        } catch (SQLException e) {
+            throw new MigrationException("Failed to update database", e);
+        }
+
+        project.getProjectProperties().setIndexedDate(Instant.now());
+        ProjectPropertiesSerialization.write(project);
+    }
+
+    private void assertObjectsExported(CdmIndexOptions options) {
```
bbpennel, line 403: Since this isn't checking that objects are exported anymore, maybe `assertCsvImportValid` or `assertCsvImportExists` or something

src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
```diff
@@ -53,8 +56,11 @@ public void setup() throws Exception {
         Files.createDirectories(project.getExportPath());
 
         fieldService = new CdmFieldService();
+        exportObjectsService = new ExportObjectsService();
```
bbpennel, line 59: Can remove

src/test/resources/files/cdm_export_fields.csv
bbpennel, line 1: I don't think this file is used anymore

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java
```diff
@@ -310,19 +305,15 @@ private void indexObject(Connection conn, List<String> exportFieldValues)
...
 
-        if (options.getCsvFile() != null) {
-            CdmFieldInfo fieldInfo = fieldService.loadFromCsvFieldsFromProject(project);
-            exportFields = fieldInfo.listAllExportFields();
-        } else {
-            CdmFieldInfo fieldInfo = fieldService.loadFieldsFromProject(project);
-            exportFields = fieldInfo.listAllExportFields();
+        CdmFieldInfo fieldInfo = fieldService.loadFieldsFromProject(project);
+        List<String> exportFields = fieldInfo.listAllExportFields();
+        if (options.getCsvFile() == null) {
```
bbpennel, line 316: I think we can remove this conditional, I think we want to have MIGRATION_FIELDS exist for CSV projects, some of the fields are used for features like group mapping.

bbpennel (changes requested):

bbpennel (approved):

PR #106, opened 2024-09-03
BXC-4676 - Service for transferring files to remote destination

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4676

* Add service for transferring source files to a remote destination, as well as create any necessary directories. 
* Updates TestSshServer to work with key authentication, and to support both scp and ssh commands
* Add SshClientService, which pulls out the functionality for running scp and ssh commands into their own class


krwong (approved):
Looks good!

PR #108, opened 2024-09-19
BXC-4677 - Add job for performing velocicroptor action

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4677

* Add job for performing velocicroptor action
* there is not a command for executing it yet, that is BXC-4689, so this is not manually testable yet


krwong (approved):
Looks good!

PR #109, opened 2024-09-23
BXC-4719 archive project

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4719](https://unclibrary.atlassian.net/browse/BXC-4719)

- add `archive` service and command to archive a list of chompb projects, which moves said projects to an "archived" folder
- modify `list_projects` service and command to count the number of archived projects


src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchiveProjectService.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Service for archiving chompb project(s)
+ * @author krwong
+ */
+public class ArchiveProjectService {
+
+    /**
+     * Archive a list of projects
+     */
+    public void archiveProject(Path currentDirectory, List<Path> projectPaths) throws IOException {
```
bbpennel, line 20: A few suggestions related to this method:
1. At that moment i think you could use this method to move any directory in the filesystem into the archived folder. It might make sense to change projectPaths to projectName, and just pass in the name of the directory for the project within the currentDirectory.
2. A user could create a project with the word "archived" in the path (we do work with library archives), which could incorrectly be labeled as archived by the status command. It would probably be safer to name the directory `.archived` so it's treated as a hidden directory and less likely to have a name collision. I'd also suggest requiring the path have `/.archive/` in it to be counted.
3. We should probably make a constant out of the name of the archive directory, and use that constant in any other places that need to reference the directory
4. The method should be `archiveProjects` since it archives multiple projects

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchiveProjectService.java
```diff
@@ -0,0 +1,32 @@
...
+        for (Path projectPath : projectPaths) {
+            if (Files.notExists(projectPath)) {
+                throw new InvalidProjectStateException("Migration project " + projectPath + " does not exist");
+            }
+
+            if (Files.isDirectory(projectPath)) {
```
bbpennel, line 26: It would probably make sense to also throw an error if the supplied value is not a directory rather than skipping it

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ListProjectsService.java
```diff
@@ -77,8 +79,14 @@ public JsonNode listProjects(Path directory) throws Exception {
                 }
             }
         }
-        log.debug(arrayNode.toString());
 
+        // count number of archived projects
+        int numberArchivedProjects = countArchivedProjects(directory);
+        ObjectNode objectNode = mapper.createObjectNode();
+        objectNode.put(ARCHIVED_PROJECTS, numberArchivedProjects);
+        arrayNode.add(objectNode);
```
bbpennel, line 87: Hm, i think it's going to be confusing if the result from this service is a list of projects, and then one of the entries in the list is just a count archived projects. Will probably also mess up the table in the admin UI. Maybe we can add this count to one of the status commands instead?

I'm also wondering if this service should know how to list archived projects in the .archived subdirectory, and include them in the regular results. Otherwise it seems like we'll never have any results with status = archived. It'd probably just mean making lines 60-81 into a helper method and then calling it on both the base directory and the archived subdirectory. If we do that, we might want to add an option so that archived projects only get included if the option is set, like `--include-archived`.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ArchiveProjectsService.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final Logger log = LoggerFactory.getLogger(ArchiveProjectsService.class);
+    public static final String ARCHIVED = ".archived";
+
+    /**
+     * Archive a list of projects
+     */
+    public void archiveProjects(Path currentDirectory, List<String> projectNames) throws IOException {
+        for (String projectName : projectNames) {
+            Path projectDirectory = currentDirectory.resolve(projectName);
+            Path archiveDirectory = currentDirectory.resolve(ARCHIVED);
```
bbpennel, line 27: this doesn't need to happen in the loop

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ListProjectsService.java
```diff
@@ -37,14 +37,15 @@ public class ListProjectsService {
     public static final String PROJECT_PATH = "projectPath";
     public static final String STATUS = "status";
     public static final String ALLOWED_ACTIONS = "allowedActions";
+    public static final String ARCHIVED_PROJECTS = "numberArchivedProjects";
```
bbpennel, line 40: I don't think this is used

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ListProjectsService.java
```diff
@@ -55,6 +56,19 @@ public JsonNode listProjects(Path directory) throws Exception {
         mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
         ArrayNode arrayNode = mapper.createArrayNode();
 
+        // list projects
+        listProjects(directory, mapper, arrayNode);
```
bbpennel, line 60: We should give this private method a different name to make it clear that its now the same method as the current one, I was wondering how it wasn't infinitely recursing at first. Maybe `listProjectsInDirectory`?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ListProjectsService.java
```diff
@@ -90,11 +101,9 @@ public JsonNode listProjects(Path directory) throws Exception {
     private String status(MigrationProject project) {
         String status = null;
 
-//        TODO: add archived state
-//        if () {
-//            status = "archived";
-//        }
-        if (!project.getProjectProperties().getSipsSubmitted().isEmpty()) {
+        if (project.getProjectPath().toString().toLowerCase().contains(ArchiveProjectsService.ARCHIVED)) {
```
bbpennel, line 104: While its unlikely to happen, I think we should check for `/.archived/` to appear in the string, to make sure we don't match something like `/path/to/gilmer.archived` or similar. 

It also doesn't need to be lowercased, since paths are case sensitive, so if it were `.ARCHIVED` we wouldn't list any projects in the directory anyways.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/services/ListProjectsService.java
bbpennel: yes you're right, it wouldn't make sense in the status commands, so removing the stat makes sense

bbpennel (approved):

PR #114, opened 2024-11-06
BXC-4768 ignore pdf children

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4768](https://unclibrary.atlassian.net/browse/BXC-4768)

- detect when object is `Document-PDF` and skip over its children
- set `ENTRY_TYPE_FIELD` / `cdm2bxc_entry_type` to `doc-pdf`

### DIFF
--- a/src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
+++ b/src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
@@ -507,6 +507,74 @@ public class CdmIndexServiceTest {
     }
 
     @Test
+    public void indexExportWithPdfCompoundObjectsTest() throws Exception {
+        Files.copy(Paths.get("src/test/resources/descriptions/pdf/index/description/desc.all"),
+                CdmFileRetrievalService.getDescAllPath(project));
+        Files.createDirectories(CdmFileRetrievalService.getExportedCpdsPath(project));
+        Files.copy(Paths.get("src/test/resources/descriptions/pdf/image/17941.cpd"),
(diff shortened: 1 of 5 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexServiceTest.java
```diff
@@ -506,6 +506,74 @@ public void indexExportWithMonographCompoundObjectsTest() throws Exception {
...
+        allFields.addAll(CdmIndexService.MIGRATION_FIELDS);
+
+        Connection conn = service.openDbConnection();
+        try {
+            Statement stmt = conn.createStatement();
+            var joinedFields = "\"" + String.join("\",\"", allFields) + "\"";
+            ResultSet rs = stmt.executeQuery("select " + joinedFields
+                    + " from " + CdmIndexService.TB_NAME + " order by " + CdmFieldInfo.CDM_ID + " asc");
+            rs.next();
+            assertEquals(17926, rs.getInt(CdmFieldInfo.CDM_ID));
```
bbpennel, line 538: It seems like it isn't actually ignoring the children of the document pdf, since "page 1" is one of them? My impression is we would not want any of the pages that appear in the cpd file to be in the index, or that we would have to ignore those pages somehow in other steps. So it might be that we need to do something like we do for compounds here:
https://github.com/UNC-Libraries/chompb/blob/c9a64a83fe300640626a116f9772e0a070ebef6e/src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmIndexService.java#L262
but instead of assigning each child a parent, we delete the child

bbpennel (approved):

PR #115, opened 2024-11-15
BXC-4769 export pdfs from cdm

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4769](https://unclibrary.atlassian.net/browse/BXC-4769)

Configure chompb to export unmapped pdf source files

- `CdmExportFilesService`: modify `exportUnmapped`  to retrieve pdf and other cpd objects from different locations, add helper method `retrieveEntryTypeField`
- add `DOWNLOADING_PDF` to `ExportState`, `ExportProgressState`, `ExportServiceState`, and `CdmExportState`
- `CdmFileRetrievalService`: add `downloadPdfFiles` method and pdf path related constants
- add/modify tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportFilesService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportFilesService.java
@@ -84,7 +84,8 @@ public class CdmExportFilesService {
                         String filePath;
                         // Pdf and image cpd objects are located in different places
                         if (CdmIndexService.ENTRY_TYPE_DOCUMENT_PDF.equals(entryTypeField)) {
-                            filename = "index.pdf";
+                            // add cdmid to filename to prevent overwriting
+                            filename = origMapping.getCdmId() + "_index.pdf";
                             filePath = pdfDir.resolve(origMapping.getCdmId() + "/index.pdf").toString();
                         } else {
                             filename = fileInfo.get(0);
--- a/src/test/java/edu/unc/lib/boxc/migration/cdm/ExportUnmappedSourceFilesIT.java
+++ b/src/test/java/edu/unc/lib/boxc/migration/cdm/ExportUnmappedSourceFilesIT.java
@@ -128,7 +128,7 @@ public class ExportUnmappedSourceFilesIT extends AbstractCommandIT {
         var sourceFileService = testHelper.getSourceFileService();
         var mappingInfo = sourceFileService.loadMappings();
         var mapping1 = mappingInfo.getMappingByCdmId("17940");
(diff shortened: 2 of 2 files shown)


src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
```diff
@@ -573,6 +573,30 @@ public void statusUnmappedVerboseTest() throws Exception {
...
+        executeExpectSuccess(args);
+
+        String[] args2 = new String[] {
+                "-w", project.getProjectPath().toString(),
+                "source_files", "status",
+                "-v" };
+        executeExpectSuccess(args2);
+
+        assertOutputMatches(".*Last Updated: +[0-9\\-T:]+.*");
+        assertOutputMatches(".*Previous Files Mapped: +0.*");
```
bbpennel, line 593: Should it be 0 across the board? I know all the file objects get dropped, but I guess it doesn't detect the pdf object itself as a file that needs to be mapped?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportFilesService.java
```diff
@@ -76,8 +77,17 @@ public String exportUnmapped() throws IOException {
...
+                        String filePath;
+                        // Pdf and image cpd objects are located in different places
+                        if (retrieveEntryTypeField(dbConn, origMapping) != null
+                                && retrieveEntryTypeField(dbConn, origMapping)
+                                .equals(CdmIndexService.ENTRY_TYPE_DOCUMENT_PDF)) {
+                            filePath = pdfDir.resolve(origMapping.getCdmId() + "/index.pdf").toString();
+                        } else {
+                            filePath = imageDir.resolve(filename).toString();
+                        }
                         var destPath = exportSourceFilesPath.resolve(filename);
```
bbpennel, line 89: What does the filename look like for these pdf files? Was just wondering if they are cpd extensions or pdf, and if it might be weird if we download the index.pdf file as a cpd file

src/test/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommandIT.java
krwong: I forgot to add `ENTRY_TYPE_DOCUMENT_PDF` to the `StatusQueryService.countIndexedFileObjects` query. Now it should say there's one file in the project and one unmapped object.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/CdmExportFilesService.java
krwong: I think the pdf filenames are index.pdf. I'll fix it.

bbpennel (approved):

PR #116, opened 2024-12-06
BXC-4808: Improve transfer reliaability

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4808

* Allow a much longer time for ssh timeouts, but keep timeout the same for auth
* Keep sessions alive for duration of thread, rather than a new one for every transfer. 
* Create all of the parent directories up front, rather than scattered throughout file transfers, and avoid extra calls. 
* Use futures so that we can get the error back from transfers and have the job fail rather than progress is a file doesn't transfer

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
@@ -7,9 +7,15 @@ import org.slf4j.Logger;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
 import static org.slf4j.LoggerFactory.getLogger;
@@ -23,6 +29,7 @@ public class SourceFilesToRemoteService {
     private SourceFileService sourceFileService;
     private SshClientService sshClientService;
     private int concurrentTransfers = 5;
+    private ExecutorService executor;
 
     /**
      * Transfer files from the source CDM server to the remote destination.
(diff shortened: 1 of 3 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
```diff
@@ -31,58 +38,90 @@ public class SourceFilesToRemoteService {
...
+            // Capture futures for transfer threads so we can receive errors
+            var futures = new ArrayList<Future<?>>();
+            for (int i = 0; i < concurrentTransfers; i++) {
+                futures.add(executor.submit(createTransferTask(pathsDeque, destinationBasePath)));
+            }
+            for (Future<?> future : futures) {
+                try {
+                    future.get();
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
```
sharonluong, line 63: would this just keep throwing exceptions as it loops through the futures?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
```diff
@@ -31,58 +38,90 @@ public class SourceFilesToRemoteService {
...
+                result.add(current);
+            }
+        }
+        return result;
+    }
+
+    private void createParentPaths(List<Path> paths, Path destinationBasePath) {
+        sshClientService.executeSshBlock((sshClient) -> {
+            for (Path path : paths) {
+                var pathRelative = path.toAbsolutePath().toString().substring(1);
```
sharonluong, line 108: what is the substring doing here? just taking off the first character?

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
bbpennel: It would throw the first exception and then the whole thing would end

src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFilesToRemoteService.java
bbpennel: Yes, it's removing the first character, so "/path/to/file" becomes "path/to/file" and then we can evaluate it as a relative path against the base path. If we evaluated "/path/to/file" against "/another/place" we'd just end up with "/path/to/file" since its absolute, but we want "another/place/path/to/file"

sharonluong (approved):

PR #118, opened 2025-01-07
BXC-4787 chompb alt-text support

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4787](https://unclibrary.atlassian.net/browse/BXC-4787)

- `alt_text_files` service and command to generate the alt-text file mapping template, validate the CSV, and get the status
- add alt-text files to sips


src/main/java/edu/unc/lib/boxc/migration/cdm/services/sips/WorkGenerator.java
```diff
@@ -205,6 +207,17 @@ protected PID addFileObject(String cdmId, String cdmFileCreated, SourceFilesInfo
             }
         }
 
+        // Link alt text file
+        if (altTextFilesInfo != null) {
+            SourceFilesInfo.SourceFileMapping altTextMapping = altTextFilesInfo.getMappingByCdmId(cdmId);
+            if (altTextMapping != null && altTextMapping.getSourcePaths() != null) {
+                Resource altTextResc = DepositModelHelpers.addDatastream(fileObjResc, DatastreamType.ALT_TEXT);
+                altTextResc.addLiteral(CdrDeposit.stagingLocation, altTextMapping.getFirstSourcePath().toUri().toString());
```
bbpennel, line 215: I checked in with Anna, and the expected structure of the CSV is: `cdmid,alt-text body`. So the mapping file won't actually contain mappings to txt files, it will contain mappings between the cdm record and the actual alt text content. Also, we need to make some adjustments so that the SIP matches what boxc expects to receive, which is for the file to be at `<sip_dir>/altText/<uuid_of_file_obj>.txt`. So it's more like how MODS work rather than like how accessFiles work.

So my suggested workflow during sip generation is:
1. Get the alt text from the mapping if it exists
2. Compute the path to the alt text file, using the uuid of the file object. See `copyDescriptionToSip`. 
3. Write the alt text to that path

For step 2, you could do it now, but I think I should update boxc to add a helper method like there is for getting the mods path. I'll work on that now.

Sorry we didn't get this all hashed out before you worked on it, I forgot to followup up with you both about expectations.

bbpennel (changes requested):

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SipServiceTest.java
```diff
@@ -1630,4 +1633,14 @@ private String convertResourcePropertiesToString(Resource resource) {
                 .map(Statement::toString)
                 .collect(Collectors.joining("\n"));
     }
+
+    private String mappingBody(String... rows) {
```
bbpennel, line 1637: Since the SipServiceTest has a lot of mappings files going on, I'd suggest giving this a scoped name like `altTextMappingBody`

bbpennel (changes requested):

bbpennel (approved):

PR #122, opened 2025-02-13
BXC-4872 add field and hookid reports for finding aid report command

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4872](https://unclibrary.atlassian.net/browse/BXC-4872)

- `finding_aid_reports` command with `field_report` to generate report for a given field and `hookid_report` to generate hookid report
- add `-f, --field` option to `FindingAidReportOptions`
- add `fieldCountUniqueValuesReport` and `hookIdReport` method to `FindingAidReportService`
- add tests


src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
```diff
@@ -0,0 +1,123 @@
...
+     * Generate the hookid report
+     * Combines values from descri and contri fields (descri_contri) and lists counts
+     * @throws Exception
+     */
+    public void hookIdReport() throws Exception {
+        String query = "select " + FindingAidService.DESCRI_FIELD + "," + FindingAidService.CONTRI_FIELD + ", count(*)"
+                + " from " + CdmIndexService.TB_NAME
+                + " where " + FindingAidService.DESCRI_FIELD + " is not null "
+                + " and " + FindingAidService.CONTRI_FIELD + " is not null "
+                + " order by count(*) desc";
```
bbpennel, line 74: I think we want to group by the fields? So something like `" group by " + FindingAidService.DESCRI_FIELD + ", " + FindingAidService.CONTRI_FIELD`.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportServiceTest.java
```diff
@@ -0,0 +1,113 @@
...
+    public void hookIdReportTest() throws Exception {
+        testHelper.indexExportData("03883");
+
+        service.hookIdReport();
+
+        assertTrue(Files.exists(service.getHookIdReportPath()));
+        try (CSVParser csvParser = parser(service.getHookIdReportPath())) {
+            List<CSVRecord> rows = csvParser.getRecords();
+            assertIterableEquals(Arrays.asList("hookid", "count"), rows.get(0));
+            assertIterableEquals(Arrays.asList("03883_folder_5", "2"), rows.get(1));
```
bbpennel, line 102: It looks like the 03883 collection only has 2 records and they both should have different hook ids, since they have contri values `folder_5` and `folder_9`, so I think the assertion here is not correct at the moment. I think it relates to my suggest back up in the FindingAidReportService and grouping. I would suggest extending 03883 so that there are actually two entries with the same hookid, and at least one entry with a different hookid so that we can verify it differentiates

bbpennel (approved):

PR #123, opened 2025-02-21
BXC-3508 chompb finding aid report command

krwong:
[https://unclibrary.atlassian.net/browse/BXC-3508](https://unclibrary.atlassian.net/browse/BXC-3508)

- add `-fa, --finding-aids` option to `generate_field_url_report` to generate a finding aids specific url report
- add `collection_report` command to `finding_aid_reports` to generate report to assess collection for associations with finding aids
- add `URL_QUERY` and `FINDING_AID_URL_QUERY` constants and `-finding-aids` option to `FieldUrlAssessmentService`
- add method `collectionReport` and helper methods `listCollectionIds`, `countRecords`, `listPotentialCollectionIds` to `FindingAidReportService`
- add and update tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
@@ -83,8 +83,8 @@ public class FindingAidReportService {
     public void hookIdReport() throws Exception {
         String query = "select " + FindingAidService.DESCRI_FIELD + "," + FindingAidService.CONTRI_FIELD + ", count(*)"
                 + " from " + CdmIndexService.TB_NAME
-                + " where " + FindingAidService.DESCRI_FIELD + " is not null "
-                + " and " + FindingAidService.CONTRI_FIELD + " is not null "
+                + " where " + FindingAidService.DESCRI_FIELD + " != ''"
+                + " and " + FindingAidService.CONTRI_FIELD + " != ''"
                 + " group by " + FindingAidService.DESCRI_FIELD + ", " + FindingAidService.CONTRI_FIELD
                 + " order by count(*) desc";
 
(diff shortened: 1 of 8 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
```diff
@@ -201,34 +202,30 @@ private Integer countRecords(String query) {
...
-                ResultSet rs = stmt.executeQuery(" select '" + field + "' from " + CdmIndexService.TB_NAME
-                        + " where '" + field + "' like " + "'^[A-Za-z0-9]{5}-?z?$'");
+                ResultSet rs = stmt.executeQuery(" select " + field + " from " + CdmIndexService.TB_NAME
+                        + " where " + field + " like " + "'^[A-Za-z0-9]{5}-?z?$' limit 10");
                 while (rs.next()) {
-                    if (!rs.getString(1).isBlank() && i < 10) {
-                        potentialCollectionIds.add(rs.getString(1));
-                        i++;
-                    }
+                    potentialCollectionIds.put(field, rs.getString(1));
```
bbpennel, line 220: I'm not sure if this is intentional, but now we will only get one value per field (whatever one is found last). We could pass back `Map<String, List<String>>` so that all the values get returned. Or keep adding the values to the existing string for that field with some kind of separator like a newline or ", ", but that might be better left up to the presentation code. If it's okay to only return one value, then the limit should probably be adjusted to 1

src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
```diff
@@ -201,34 +202,30 @@ private Integer countRecords(String query) {
...
                 }
             }
         } catch (SQLException e) {
             throw new MigrationException("Error interacting with export index", e);
         }
 
         if (potentialCollectionIds.isEmpty()) {
-            potentialCollectionIds.add("no potential collection ids found");
```
bbpennel: Since there's going to be an `if` either way, maybe it'd make sense to just pass back an empty potentialCollectionIds and then in `collectionReport` have an `if` that checks if the result is empty and displays this message rather than calling `showFieldMapValues`? That would help separate the concerns of generating the list of matches from the presentation of results.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/FindingAidReportService.java
```diff
@@ -102,6 +117,146 @@ public Path getHookIdReportPath() {
...
+        try (Connection conn = indexService.openDbConnection()) {
+            var stmt = conn.createStatement();
+            for (String field : exportFields) {
+                List<String> ids = new ArrayList<>();
+                ResultSet rs = stmt.executeQuery(" select " + field + " from " + CdmIndexService.TB_NAME
+                        + " where " + field + " like " + "'^[A-Za-z0-9]{5}-?z?$' limit 10");
+                while (rs.next()) {
+                    ids.add(rs.getString(1));
+                }
+                potentialCollectionIds.put(field, ids);
```
bbpennel, line 227: I think there might need to be a check to see if ids is empty, otherwise the map will never be empty

PR #124, opened 2025-03-10
BXC-4843 boxctron command accept exclusions

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4843](https://unclibrary.atlassian.net/browse/BXC-4843)

- add `-e, --exclusion-csv` option
- add `EXCLUSIONS_CSV_HEADERS` and rename `DATA_CSV_HEADERS` to `BOXCTRON_CSV_HEADERS`
- add `gatherExclusionPaths` helper method
- add tests
- update build.yml actions to v3 and v4 (or build fails)

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/BoxctronFileService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/BoxctronFileService.java
@@ -17,7 +17,9 @@ import java.nio.file.Path;
 import java.sql.Connection;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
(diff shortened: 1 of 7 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/BoxctronFileService.java
```diff
@@ -47,8 +49,13 @@ public void generateMapping(BoxctronFileMappingOptions options) throws IOExcepti
         assertProjectStateValid();
         ensureMappingState(options);
 
-        // Gather listing of all potential source file paths to match against
+        // Gather list of all potential source file paths to match against
         List<String> candidatePaths = gatherCandidatePaths(getVelocicroptorDataPath(project.getProjectPath()));
+        // Gather list of all source file paths to exclude
+        List<String> exclusionPaths = new ArrayList<>();
+        if (options.getExclusionsCsv() != null && Files.exists(options.getExclusionsCsv())) {
```
bbpennel, line 56: I'd suggest letting it fail if the file doesn't exist rather than proceeding like no file was provided, so that the user will be aware they provided an invalid path.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/BoxctronFileService.java
```diff
@@ -47,8 +49,13 @@ public void generateMapping(BoxctronFileMappingOptions options) throws IOExcepti
         assertProjectStateValid();
         ensureMappingState(options);
 
-        // Gather listing of all potential source file paths to match against
+        // Gather list of all potential source file paths to match against
         List<String> candidatePaths = gatherCandidatePaths(getVelocicroptorDataPath(project.getProjectPath()));
+        // Gather list of all source file paths to exclude
+        List<String> exclusionPaths = new ArrayList<>();
```
bbpennel, line 55: Just a minor efficiency note, since this list is exclusively for lookups, it would be more efficient to use a Set/HashSet since the time for `contains` calls is constant, meaning it won't take any more time to search a set of 10 or 1000000 entries, versus a list where the search time depends on the number of entries. If you're curious https://www.baeldung.com/java-collections-complexity

bbpennel (changes requested):

bbpennel (approved):

PR #127, opened 2025-05-21
BXC-4945 workflow to import ref ids

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4945](https://unclibrary.atlassian.net/browse/BXC-4945)

- add `AspaceRefIdService` and `AspaceRefIdInfo`
- add `ASPACE_REF_ID_MAPPING_FILENAME` to `MigrationProject`, `aspaceRefIdMappingsUpdatedDate` and `aspaceRefIdMappingsSyncedDate` to `MigrationProjectProperties`, and `ASPACE_REF_ID` column to index, 
- `aspace_ref_id generate` command to generate `ref_id_mapping.csv` template with work ids populated and an empty aspace ref id column
- `aspace_ref_id sync` command to import `ref_id_mapping.csv` into the index
- update SIP generation to add aspace ref ids to works if said ref ids are populated in the database
- add tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
@@ -1,0 +1,218 @@
+package edu.unc.lib.boxc.migration.cdm.services;
+
+import edu.unc.lib.boxc.migration.cdm.exceptions.InvalidProjectStateException;
+import edu.unc.lib.boxc.migration.cdm.exceptions.MigrationException;
+import edu.unc.lib.boxc.migration.cdm.model.AspaceRefIdInfo;
+import edu.unc.lib.boxc.migration.cdm.model.CdmFieldInfo;
+import edu.unc.lib.boxc.migration.cdm.model.MigrationProject;
+import edu.unc.lib.boxc.migration.cdm.util.ProjectPropertiesSerialization;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVPrinter;
+import org.apache.commons.csv.CSVRecord;
+import org.slf4j.Logger;
+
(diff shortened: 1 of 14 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
```diff
@@ -0,0 +1,218 @@
...
+        project.getProjectProperties().setAspaceRefIdMappingsSyncedDate(timestamp);
+        ProjectPropertiesSerialization.write(project);
+    }
+
+    public Path getMappingPath() {
+        return project.getAspaceRefIdMappingPath();
+    }
+
+    private List<String> getIds() {
+        if (ids == null) {
```
bbpennel, line 127: I'd suggest not using an instance variable for `ids`, it means that the service can only be used to load ids once. If the method were going to be called a lot of times it'd make sense to cache the value, but i think it's only being used once, so it probably makes more sense to keep the service stateless

src/test/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdServiceTest.java
```diff
@@ -0,0 +1,284 @@
...
+        groupOptions.setGroupFields(Arrays.asList("groupa"));
+        GroupMappingService groupService = testHelper.getGroupMappingService();
+        groupService.generateMapping(groupOptions);
+        groupService.syncMappings(makeDefaultSyncOptions());
+    }
+
+    private void assertWorkSynced(Connection conn, String expectedCdmId, String expectedAspaceRefId)
+            throws Exception {
+        Statement stmt = conn.createStatement();
+        ResultSet rs = stmt.executeQuery("select * from " + CdmIndexService.TB_NAME
```
bbpennel, line 255: This is probably not working either

bbpennel (approved):

PR #129, opened 2025-06-19
BXC-4967 generate aspace ref id mapping from hook id lookups

krwong:
[https://unclibrary.atlassian.net/browse/BXC-4967](https://unclibrary.atlassian.net/browse/BXC-4967)

- `AspaceRefIdCommand`: add `generate_from_csv` command to generate aspace ref id mapping using the `hookid_to_refid_map.csv` file
- `AspaceRefIdService: add constants, rename `generateAspaceRefIdMapping` to `generateBlankAspaceRefIdMapping`, add generateAspaceRefIdMappingFromHookIdRefIdCsv and helper methods
- add tests

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
@@ -20,7 +20,9 @@ import java.nio.file.Path;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
```diff
@@ -71,6 +73,10 @@ public void generateBlankAspaceRefIdMapping() throws IOException {
     public void generateAspaceRefIdMappingFromHookIdRefIdCsv() throws IOException {
         assertProjectStateValid();
 
+        if (!hasProjectContriAndDescriFields()) {
```
bbpennel, line 76: We should add a test to verify that this works. Can probably use the "monograph" that we already have sitting around in the test resources.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
```diff
@@ -55,10 +54,18 @@ public AspaceRefIdService() {
     public void generateBlankAspaceRefIdMapping() throws IOException {
         assertProjectStateValid();
 
+        if (!hasProjectContriAndDescriFields()) {
```
bbpennel, line 57: hmm, i don't know if this is desirable, you might want to check in with anna. This would mean we wouldn't be able to generate blank mappings for non-cdm collections without adding in vestigial fields to the project. It would probably be fine to have an empt hookid column in the csv file. 

I know it's awkward dealing with fields that may or may not be there in sql tables. A more complicated option would be to have it run a different getIdsAndHookIds query that doesn't retrieve to the hookid field if `hasProjectContriAndDescriFields` returns false (it'd just return "" in the select clause), but I don't know if that's essential.

src/main/java/edu/unc/lib/boxc/migration/cdm/services/AspaceRefIdService.java
```diff
@@ -187,12 +169,33 @@ private Map<String, String> getHookIdsAndRefIds(Path mappingPath) throws IOExcep
...
+                } else if (key.equalsIgnoreCase(FindingAidService.DESCRI_FIELD)
+                        && value.equalsIgnoreCase(FindingAidService.COLLECTION_NUMBER_FIELD_DESC)) {
+                    hasDescri = true;
+                }
+            }
+
+            if (hasContri && hasDescri) {
+                projectHasContriAndDescri = true;
+            } else {
+                projectHasContriAndDescri = false;
```
bbpennel, line 197: could be shortened to `projectHasContriAndDescri = hasContri && hasDescri;`

PR #130, opened 2025-06-27
BXC-4981 - Fix too many file handles during description generate command

bbpennel:
https://unclibrary.atlassian.net/browse/BXC-4981

* Close streams during description expand operation to fix issue of running out of handles in a large collection.
* Add a unit test for the generate command


krwong (approved):
Looks good!

PR #150, opened 2026-04-06
Bxc 5577 disk usage

sharonluong:
https://unclibrary.atlassian.net/browse/BXC-5577

### DIFF
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
@@ -170,6 +170,7 @@ public class SourceFilesCommand {
                     "project’s source_files.csv")
     public int calculateDiskUsage() {
         try {
+            initialize(false);
             var diskUsage = sourceService.calculateStorage();
             outputLogger.info("Source files total " + diskUsage + "MB in storage");
             return 0;
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
@@ -170,6 +170,8 @@ public class SourceFilesCommand {
                     "project’s source_files.csv")
     public int calculateDiskUsage() {
         try {
+            var diskUsage = sourceService.calculateStorage();
+            outputLogger.info("Source files total " + diskUsage + "MB in storage");
             return 0;
         } catch (Exception e) {
             log.error("Disk usage calculation failed", e);
--- a/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
+++ b/src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
@@ -172,11 +172,11 @@ public class SourceFilesCommand {
         try {
             initialize(false);
             var diskUsage = sourceService.calculateStorage();
-            outputLogger.info("Source files total " + diskUsage + "MB in storage");
+            outputLogger.info("Source files total {} MB in storage", diskUsage);
             return 0;
         } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


src/main/java/edu/unc/lib/boxc/migration/cdm/services/SourceFileService.java
```diff
@@ -557,6 +558,25 @@ public static CSVPrinter openMappingsPrinter(Path mappingPath) throws IOExceptio
...
+     * @return total length in MB
+     */
+    public long calculateStorage() throws IOException {
+        long totalBytes = 0;
+        try (var csvParser = openMappingsParser(project.getSourceFilesMappingPath())) {
+            for (CSVRecord record : csvParser) {
+                var basePath = Paths.get(record.get(SOURCE_FILE_FIELD));
+                var file = basePath.toFile();
+                var length = file.length();
+                totalBytes += file.length();
```
bbpennel: A few things here:
* Calling file.length() twice. 
* we can skip converting to a file object by using `Files.size` which does the same thing but for Paths.
* The SOURCE_FILE_FIELD isn't always going to be populated, so we should check that its empty before calculating sizes.

src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
```diff
@@ -165,6 +165,22 @@ public int add(@Mixin AddSourceFileMappingOptions options) throws Exception {
...
     }
 
+    @Command(name = "disk_usage",
+            description = "Calculate the disk usage of all of the source files mapped in a migration " +
+                    "project’s source_files.csv")
+    public int calculateDiskUsage() {
+        try {
+            initialize(false);
+            var diskUsage = sourceService.calculateStorage();
+            outputLogger.info("Source files total " + diskUsage + "MB in storage");
```
bbpennel: use `{}` for interpolation

src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
```diff
@@ -165,6 +165,22 @@ public int add(@Mixin AddSourceFileMappingOptions options) throws Exception {
...
+            return 0;
+        } catch (Exception e) {
+            log.error("Disk usage calculation failed", e);
+            outputLogger.info("Disk usage calculation failed: {}", e.getMessage(), e);
```
bbpennel: we'll get the message part of the exception twice here, can leave off either the `e.getMessage()` or `e` depending on if you want it to display the full stacktrace.

src/test/java/edu/unc/lib/boxc/migration/cdm/services/SourceFileServiceTest.java
```diff
@@ -763,6 +763,12 @@ public void addToMappingEmptyCsvTest() throws Exception {
         assertEquals(2, info.getMappings().size());
     }
 
+    @Test
+    public void calculateDiskStorageTest() throws IOException {
+        writeCsv(mappingBody("testid,," + filesystemSourceFile("IMG_2377.jpeg") + ","));
```
bbpennel: we should test with more than one file to verify that it is summing correctly. There is `D2_035_Varners_DrugStore_interior.tif` which is 25mb so should give a good sum.

It'd also be good to have an entry where there is an id but no source file path to verify it can handle partially populated files.

bbpennel (changes requested):

src/main/java/edu/unc/lib/boxc/migration/cdm/SourceFilesCommand.java
sharonluong: oh! i was copying what the other commands' output loggers had. i'll remove some

bbpennel (approved):

